PR #1, opened 2017-10-28
Add campaign model

chrisruenes1:
-Pushed up schema—I'm on a team at work where we do NOT commit our schema, but I think that, while the codebase remains small, it is definitely easiest if we DO (rather than relying on migrations on data dumps)

-Holding off on setting up a test framework until we decide what we want to use (I'm used to rspec, but open to other suggestions)


brycemcd:
@chrisruenes1 LGTM :shipit: 

Until we have a compelling reason _not_ to commit our schema, I think it's a good idea that we do so. Also, Rspec is my testing framework of choice as well. Go ahead and add that when you get a chance.

:+1:

chrisruenes1:
@brycemcd added rspec and a basic test for the model

spec/models/campaign.rb
```diff
@@ -0,0 +1,15 @@
+require 'rails_helper'
+
+describe Campaign do
+  describe "#name"
```
brycemcd: shouldn't this be a block?

```ruby
describe "#name" do
  it "something seomthing" do
    # ...
  end
end

File: spec/models/campaign.rb [hunk 2]
```diff
@@ -0,0 +1,15 @@
...
+  it "should validate presence" do
+    campaign = Campaign.new
+    campaign.name = ''
+    campaign.valid?
+    expect(campaign.errors[:name]).to include("can't be blank")
+    
+    campaign.name = 'test1'
+    campaign.valid?
+    expect(campaign.errors[:name]).not_to include("can't be blank")
+  end
```
brycemcd: This is a lot of code for testing a simple validation (a case _could_ be made that it's testing rails and not the model but I think this is worthwhile). If it's easier for you, load (should matchers)[https://github.com/thoughtbot/shoulda-matchers] into the dependencies and replace this whole block with `it { should validate_presence_of(:name) }`. It's your call.

PR #2, opened 2017-12-22
Adds Heroku config

brycemcd:
## What

The simplest heroku config that could possibly work

## How to Test

After we have a working app, we can more appropriately test


chrisruenes1:
👍

PR #5, opened 2018-01-12
Bill summaries

chrisruenes1:
NOTE: DO NOT MERGE THIS IN BEFORE MERGING IN PR#4, ENV SETUP

This PR sets up the rails api and builds the backend for bill summaries. It also sets up the frontend, including actions, reducers, and components for a BillsIndex and a Bills Add/Edit form.

To test: 

- [ ]  start the app. It should start without problems

- [ ] test all the crud functionality (you may need to add a campaign via the rails console first (just `Campaign.create(name: "campaign1")` will do

- [ ] make sure there are no strange navigation issues (forward and back buttons work, as do simply entering urls)

- [ ] run the tests!


frontend/actions/billActions.js
```diff
@@ -0,0 +1,106 @@
...
+      error: (data) => {
+        // TODO build in error handling from backend
+        console.log(data);
+      },
+    });
+  };
+}
+
+export function destroyBill(billId) {
+  return function destroyBillTunk(dispatch) {
```
e91096: tunk!

e91096:
did you branch off your env set up? if so, can i switch the base branch for this pr--just for code review?

chrisruenes1:
Esther yes and please feel free to switch the base branch!

app/controllers/api/bill_summaries_controller.rb
```diff
@@ -0,0 +1,45 @@
+class Api::BillSummariesController < ApplicationController
```
e91096: i wonder if it'll ever get confusing down the line that we refer to these entities as bills on the front end and bill summaries on the backend?

db/migrate/20180112044042_create_bill_summary.rb
```diff
@@ -0,0 +1,11 @@
+class CreateBillSummary < ActiveRecord::Migration[5.1]
+  def change
+    create_table :bill_summaries do |t|
+      t.belongs_to :campaign
```
e91096: whoa. i did not know you could do this!

frontend/components/BillsIndex.jsx
```diff
@@ -0,0 +1,88 @@
...
+      ];
+      return (
+        <CrudRow
+          actions={['edit', 'destroy']}
+          cells={cells}
+          editRootUrl="/bills"
+          handleDestroy={this.handleDestroy}
+          handleEdit={push}
+          id={bill.id}
+          key={hash(bill)}
```
brycemcd: To my own point earlier about getting routing values from the rails router, this may be a place to do that (I have no idea). Adding something like: `data-api-endpoint-path="/bills/123" data-api-endpoint-method="PATCH"` to the markup could make it accessible to the form submission code.

spec/models/bill_summary_spec.rb
```diff
@@ -0,0 +1,12 @@
...
+RSpec.describe BillSummary, type: :model do
+  subject { BillSummary.new(bill_number: 's555')}
+  it { should validate_presence_of(:bill_number) }
+  it { should validate_uniqueness_of(:bill_number) }
+
+  it { should allow_value('hr2153').for(:bill_number) }
+  it { should allow_value('s88888').for(:bill_number) }
+  it { should_not allow_value('p253').for(:bill_number) }
+  it { should_not allow_value('s122347823423').for(:bill_number) }
+end
```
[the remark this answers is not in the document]
brycemcd: It's a good point, but I think it gets into testing rails (or active record to be more precise) itself. AR tests its validators and testing the errors would be testing the tests IMHO. If there are use cases where we're using a validator with a lot of extra decorations around it, it's probably worth testing. If we're just using vanilla AR vaidators I think the contributors have tested that pretty well.

db/migrate/20180112044042_create_bill_summary.rb
```diff
@@ -0,0 +1,11 @@
+class CreateBillSummary < ActiveRecord::Migration[5.1]
+  def change
```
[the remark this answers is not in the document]
chrisruenes1: @brycemcd I'll defer to you on this, but I think that change is ok for any easily invertible operation. So with creating tables, adding/removing columns (as long as the type is present) etc, I usually just use change, but if it involves any sort of my complex logic, I'll usually write an up and down

app/controllers/api/bill_summaries_controller.rb
chrisruenes1: Will change this!

app/controllers/api/bill_summaries_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+  def index
+    bills = BillSummary.all
+    render json: bills
+  end
+  
+  def create
+    bill = BillSummary.new(permitted_params)
+    if bill.save!
```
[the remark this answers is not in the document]
chrisruenes1: you're absolutely right, thanks for the catch!

frontend/actions/billActions.js
```diff
@@ -0,0 +1,106 @@
...
+    type: 'REMOVE_BILL',
+    payload: { id: bill.id },
+  };
+}
+
+
+export function createBill(params) {
+  return function createBillThunk(dispatch) {
+    $.ajax({
+      url: '/api/bill_summaries',
```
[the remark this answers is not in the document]
chrisruenes1: @brycemcd hardcoding in the urls of ajax requests is definitely idiomatic in a redux app. I'm open to alternatives though—I've not tried to access rails router code from the the frontend before, do you have an example?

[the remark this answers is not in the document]
chrisruenes1: Oh also @brycemcd a thunk is a function that wraps another function to delay it's execution. Since action creators in redux ultimately need to return an object in the format { type: "SOME_CONSTANT", payload: { someData } }, thunk is a middleware that allows the `dispatch` to be passed to any asynchronous function defined inside an action creator, so that it can dispatch an object in that format as a callback

chrisruenes1:
I don't have time to finish these changes tonight but hope I will tomorrow!

app/controllers/api/bill_summaries_controller.rb
chrisruenes1: ok it should be bill_summary everywhere now!

chrisruenes1:
@e91096 @brycemcd I've implemented the feedback from above and rebased on master to resolve merge conflicts, if either or both of you would be willing to do one last check to make sure everything still works for you locally, I can merge this in after that!

Gemfile
```diff
@@ -20,6 +20,8 @@ gem 'uglifier', '>= 1.3.0'
 # See https://github.com/rails/execjs#readme for more supported runtimes
 # gem 'therubyracer', platforms: :ruby
 
+# Use CoffeeScript for .coffee assets and views
+gem 'coffee-rails', '~> 4.2'
```
brycemcd: PR 5: the return of coffeescript?

db/migrate/20180112044042_create_bill_summary.rb
```diff
@@ -0,0 +1,11 @@
...
+    create_table :bill_summaries do |t|
+      t.belongs_to :campaign
+      t.text :summary
+      t.string :bill_number, null: false, unique: true
+      t.string :full_bill_url
+      t.timestamps
```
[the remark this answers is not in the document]
brycemcd: bump

PR #7, opened 2018-02-05
Add new user registration form, set up devise, figaro

e91096:
[branched off bill_summaries]
*requires migration

https://trello.com/c/XeouPPRI/7-user-registration

to do:
- set `DEVISE_SECRET_KEY` on heroku (`figaro heroku:set -e production`)

not included:
- logged in state?
- fe tests
- error handling


app/controllers/api/users_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+    else
+      render json: user.errors.full_messages, status: 422
+    end
+  end
+
+  def destroy
+    user = User.find(params[:id])
+    user.destroy
+    render json: user
+  end
```
brycemcd: for now, let's remove the endpoint to destroy users

config/environments/development.rb
```diff
@@ -51,4 +51,7 @@
   # Use an evented file watcher to asynchronously detect changes in source code,
   # routes, locales, etc. This feature depends on the listen gem.
   config.file_watcher = ActiveSupport::EventedFileUpdateChecker
+
+  # for devise TODO prod config
+  config.action_mailer.default_url_options = { host: 'localhost', port: 3000 }
```
brycemcd: for prod, do you want to set up a config using env vars? I have an SMTP server we can use to get going and then make a note that we'll need an actual SMTP relay very soon.

config/initializers/devise.rb
```diff
@@ -0,0 +1,277 @@
...
+  # db field (see migrations). Until confirmed, new email is stored in
+  # unconfirmed_email column, and copied to email column on successful confirmation.
+  config.reconfirmable = true
+
+  # Defines which key will be used when confirming an account
+  # config.confirmation_keys = [:email]
+
+  # ==> Configuration for :rememberable
+  # The time the user will be remembered without asking for credentials again.
+  # config.remember_for = 2.weeks
```
brycemcd: Set this for a long time. A year?

config/routes.rb
```diff
@@ -1,7 +1,9 @@
 Rails.application.routes.draw do
 
+  devise_for :users
   namespace :api do
     resources :bill_summaries, only: [:index, :show, :create, :update, :destroy]
+    resources :users, only: [:show, :create, :update, :destroy]
```
brycemcd: please remove the `:destroy` endpoint. Scary.

db/migrate/20180115004737_create_users.rb
```diff
@@ -0,0 +1,15 @@
+class CreateUsers < ActiveRecord::Migration[5.1]
+  def change
+    create_table :users do |t|
+      t.string :congressional_district
+      t.string :email, null: false, unique: true
+      t.string :first_name
+      t.string :last_name
+      t.string :state
+      t.boolean :US_citizen
+      t.string :zip
```
brycemcd: Given that all fields are required, please set `null: false`. I should have been more clear in the Trello ticket (sorry!) but the congressional district should be a 1-3 digit integer. `:US_citizen` should be lowercased. I'm pretty sure that pg would lowercase it anyway but IMHO it's better to be consistent with that convention in the migration file as well.

app/models/user.rb
```diff
@@ -0,0 +1,8 @@
+class User < ApplicationRecord
+  # Include default devise modules. Others available are:
+  # :confirmable, :lockable, :timeoutable and :omniauthable
+  devise :database_authenticatable, :registerable,
+         :recoverable, :rememberable, :trackable, :validatable
+  # validations
+  validates :email, presence: true, uniqueness: true
```
brycemcd: Some validations on the district and state should be present here too. Districts are 1-3 digit integers and states should be two characters (letters only). We'll need a way of making that clear in the UI.

frontend/actions/user_actions.js
```diff
@@ -0,0 +1,90 @@
...
+    },
+  };
+}
+
+function removeUser(user) {
+  return {
+    type: 'REMOVE_USER',
+    payload: { id: user.id },
+  };
+}
```
brycemcd: please don't allow for removing users :)

frontend/components/RegistrationForm.jsx
```diff
@@ -0,0 +1,194 @@
...
+  validateInput = (fieldName, value) => {
+    const errors = [];
+
+    if (fieldName === 'USCitizen') return errors;
+
+    const trimmedValue = value.trim();
+    switch (fieldName) {
+      case 'congressionalDistrict':
+        if (!trimmedValue.match(/^\d{1,2}$/)) {
+          errors.push('Congressional District should be a 1 or 2 digit number.');
```
brycemcd: oh, neat! FE validation!

As a general rule, I think it's a good idea to do validation in the browser, but that's easily turned off. It's good to validate in the backend and database code (as much as is reasonable) as well.

frontend/components/RegistrationSuccess.jsx
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+const RegistrationSuccess = () => <div>You did it!</div>;
```
brycemcd: such the feeling of accomplishment

frontend/components/shared/Input.jsx
```diff
@@ -34,25 +34,44 @@ export default class Input extends Component {
...
+        );
+      case 'text':
+        return (
+          <textarea
+            {...this.getInputProps()}
+            style={{
+              fontSize: '14px',
+              width: '425px',
+              height: '100px',
+              resize: 'none',
```
brycemcd: Is this just to get it to render? I would assume this is in the CSS later?

frontend/reducers/user_reducers/byId.js
```diff
@@ -0,0 +1,21 @@
...
+export default function byId(state = {}, action) {
+  const { type, payload } = action;
+  switch (type) {
+    case 'ADD_USER':
+      return {
+        ...state,
+        [payload.id]: payload.user,
+      };
+    case 'REMOVE_USER':
+      return _.omit(state, payload.id);
```
brycemcd: that same ol' delete/destroy/remove user comment again

PR #8, opened 2018-02-10
Add campaigns crud

chrisruenes1:
This is the added stuff for the campaign form. I added a few missing links between the data. It should be fairly self-explanatory, and my computer is dying :)


frontend/components/CampaignsIndex.jsx
```diff
@@ -0,0 +1,82 @@
...
+        <table>
+          <tbody>
+            <tr>
+              <th>Name</th>
+              <th>Number of Bill Summaries</th>
+            </tr>
+            {this.renderCampaignRows()}
+          </tbody>
+        </table>
+        <a href="/campaigns/new" onClick={push}>Add Campaign</a>
```
brycemcd: Does React have a concept similar to rails' `new_campaigns_path`?

chrisruenes1: It doesn't. But that doesn't mean we couldn't implement some sort of parser that would behave similarly. Let's discuss how we could design this when we have our next hack night

PR #9, opened 2018-02-10
Add seeds

chrisruenes1:
I added some seeds, and a fairly wacky mechanism for silent failure if the model in question does not yet exist on the branch. My computer is still about to die ;)


brycemcd:
:+1:

PR #10, opened 2018-02-13
Continuous Integration / Deployment Pipeline update

brycemcd:
Automatically deploying this app when code is merged into master turned out to be a bit of a beast. The same build pipeline had to be configured in Circle CI and Heroku. (My original intention was to build it in CI and then have Heroku receive the packaged app at the end of the build.)

This PR adds webpack to production dependencies so that the build pipelines are able to compile the assets and move them to the right place.

It updates the Circle CI configuration and provides stubs for executing the unit tests of both ruby and react.

Once tests pass, Github will be notified and Heroku will then pull the app to production iff the build was done on the master branch.


brycemcd:
I got a :+1: from Esther in Slack. I'm counting it

PR #11, opened 2018-03-04
Create a rudimentary navbar that uses a constants file + push function

chrisruenes1:
This branch adds a navbar! 

The top level component is now wrapped in a Layout component. For now, this Layout just wraps the new Navbar, but we could add other global components like an error handler down the road as well. 

The navbar just iterates through a constants file that defines an array of POJOs with a `name` and a `path` property. For each of these POJOs, it creates a `NavItem` component that just handles a click by calling `push` with its `path`.

To test:

1. Make sure manual navigation by typing into the url bar still works as expected
2. Make sure the existing links on the navbar navigate properly
3. Try adding a new POJO to the constants file and verify that it shows up in the Navbar and that clicking on it redirects to the path specified

A couple other notes:
1. Merging at some point preserved both the bundle of `bills` FE stuff and `billSummaries`. I removed all `bills` stuff in this MR
2. @e91096 I think we defined a constants file on another branch right? We'll probably have some minor merge conflicts and we should figure out the best way to handle these merges
3. We should begin to prioritize writing frontend tests soon before it becomes a real shitty project to have to cover all of this stuff! @e91096 maybe we can take a long work session just to write tests, after we finish with the remaining open cards on Trello?


frontend/components/NavItem.jsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import push from '../push';
+
+export default function NavItem(props) {
+  const handleClick = () => {
+    push(props.path);
+  };
+
+  const style = props.selected ? { color: 'pink', fontSize: '24px' } : {};
```
brycemcd: shouldn't this be defined in CSS? I'm not sure if it's a convention in React to use inline styles (if it is, I don't like it :) ). I would encourage you to keep structure (HTML) separate from behavior (javascript) separate from decoration/style (CSS). I know React likes to conflate structure and behavior but for this, I would define a class that is applied to the HTML given a certain state (i.e. when the `props.selected` is true). Then, in CSS decorate the class the way you want it to look.

frontend/components/Navbar.jsx
```diff
@@ -0,0 +1,31 @@
...
+  const isItemSelected = (pathName) => {
+    const matcher = new RegExp(pathName);
+    return !_.isNull(window.location.href.match(matcher));
+  };
+
+  const navBarStyle = {
+    float: 'left',
+    minHeight: '100px',
+    paddingRight: '50px',
+  };
```
brycemcd: same comment here as above. Define a class and decorate the class in CSS.

brycemcd:
Is this still something we want to merge at some point?

chrisruenes1:
@brycemcd I actually wound up adding this to the votes fe, so we don't need this anymore! I'm gonna close this PR

PR #12, opened 2018-04-19
backend - consistently format errors sent to frontend

e91096:
https://trello.com/c/dhXwue5T/23-wip-errors-returned-to-the-frontend-should-be-consistent-objects

controllers now returns errors to the frontend as errors (i.e. using error status codes) formatted the following way:
```
{
  messages: ['error message 1', 'error messages 2']
}
```

i also replaced the `nothing: true` responses with 'not found' error messages bc it looks like `:nothing` is deprecated (https://stackoverflow.com/questions/34688726/the-nothing-option-is-deprecated-and-will-be-removed-in-rails-5-1)!

i did also read a little bit about custom error classes to abstract error handling, but i wasn't sure if this was the right time for something like that. thoughts?
https://medium.com/@stevenpetryk/providing-useful-error-responses-in-a-rails-api-24c004b31a2e
http://www.thegreatcodeadventure.com/rails-api-painless-error-handling-and-rendering-2/


brycemcd:
@e91096 are you okay if we close this now? I think we have a couple of other options in flight for formatting our responses.

PR #13, opened 2018-06-22
Dockerize Constituents

brycemcd:
# Now With More Docker
This PR sets up a development environment using docker. Hopefully, this makes it easier to onboard new people (like me) onto developing the constituent app.

Images can be pushed to [our dockerhub account](https://hub.docker.com/u/pocketlobby/)


README.md
```diff
@@ -1,21 +1,47 @@
...
-13. we use eslint with the airbnb style guide. here is a guide to set up linter eslint w/ airbnb in atom
+# Setup
+
+### Docker
+
+Images can be found under the Pocket Lobby hub.docker.com account. We're
+standardizing on using the circleci images of ruby simply to aid in
+efficiency of CI/CD.
+
+To begin, either pull from dockerhub or build it yourself using this
```
e91096: should we add the bit about dumping your db if you want to save anything in your db?

brycemcd: yeah, great idea. I'll add that and push it up

README.md
```diff
@@ -1,21 +1,47 @@
...
+command: `docker build -t pocketlobby/constituents .` That will build an
+image using the Dockerfile in this repository.
+
+Environment variables can be exposed to the docker images by adding a
+file like `docker/env` and linking to it in the `docker-compose` file.
```
e91096: this is maybe a silly question, but does `docker/env` need to have the things currently in `config/application.yml`?
also, does this change the way we access things like the psql and rails consoles?

brycemcd: So long as your config/application.yml file is included in the source you shouldn't need to add it to any docker(-compose) configuration.

If you boot up your environment with anything other than application.yml used for environment variables then _those_ variables would need to be added to the docker file or included in the command. i.e. if you boot like this `FOO=bar bundle exec rails s` you'd need to add `-e FOO=bar` to the docker command. I don't _think_ we do that though.

brycemcd:
@e91096 I was having some trouble making trivial updates. I've updated this PR. Would you mind taking one more spin through code review please?

brycemcd:
@andrezafeu those changes LGTM. I'll wait for @e91096's re-approval before merging this into master

README.md
```diff
@@ -1,21 +1,49 @@
...
+
+Bootstrap the environment:
+
+`cp docker/env.example docker/env`
+
+Change the password in the `docker/env` file if you'd like.
+
+Create a Docker image from the definition contained in the Dockerfile:
+
+`docker-build`
```
e91096: is this missing anything?

docker/README.md
```diff
@@ -0,0 +1,67 @@
+# Dockerized Constituents
+
+We use docker to manage our development, test, CI and production
+environments.
+
+To get started install docker and run this command:
+
+`docker-compose up`
```
e91096: this doesn't seem to be starting up rails for me anymore... 😕

PR #14, opened 2018-07-17
User Authentication

e91096:
[Trello](https://trello.com/c/lyGxLIdd/8-user-authentication)

- A user can now register, log in, and log out.
- A user must be logged in to access campaigns, bill summaries, and votes. (Note: the backend endpoint will fail if a user is not logged in, but the frontend route, i.e. page will still load with no indication of an error. We should talk about desired behavior.)


app/controllers/api/campaigns_controller.rb
```diff
@@ -3,7 +3,7 @@ def index
     campaigns = Campaign.all
     render json: campaigns.map { |campaign| campaign_json(campaign) }
   end
-  
+
```
brycemcd: :skull: to extra whitespace. Nice

app/controllers/api/registrations_controller.rb
```diff
@@ -0,0 +1,27 @@
+class Api::RegistrationsController < Devise::RegistrationsController
+  clear_respond_to
+  respond_to :json
+
+  def create
+    build_resource(sign_up_params)
+    resource.save
```
brycemcd: I haven't read the docs on Devise in a long time. If the save is not successful is it automatically handled? Reading this, it looks like you're assuming the params will always be valid.

This and the sessions_controller are both in the api namespace. Is that appropriate? For now, all the clients are belong to us so no third party callers will exist. It's not immediately intuitive to me that this should be part of an API per-se but rather just part of our app.

app/controllers/application_controller.rb
```diff
@@ -2,5 +2,47 @@ class ApplicationController < ActionController::Base
...
+    else
+      validation_error(resource)
+    end
+  end
+
+  def validation_error(resource)
+    render json: {
+      errors: [
+        {
+          status: '400',
```
brycemcd: I think a response status of 422 or 418 would be more appropriate for form validation errors. [Wikipedia HTTP status codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_Client_errors)

what does `code: '100'` mean?

app/controllers/application_controller.rb
```diff
@@ -2,5 +2,47 @@ class ApplicationController < ActionController::Base
...
+    devise_parameter_sanitizer.permit(
+      :sign_up,
+      keys: [
+        :congressional_district,
+        :first_name,
+        :last_name,
+        :state,
+        :us_citizen,
+        :zip
+      ]
```
brycemcd: Is this a duplication of [this](https://github.com/PocketLobby/constituents/pull/14/files#diff-dcb74582a4980f9bbaf3dcb05f5af796R13) ?

app/models/user.rb
```diff
@@ -1,8 +1,8 @@
 class User < ApplicationRecord
-  # Include default devise modules. Others available are:
-  # :confirmable, :lockable, :timeoutable and :omniauthable
   devise :database_authenticatable, :registerable,
-         :recoverable, :rememberable, :trackable, :validatable
+         :recoverable, :rememberable, :trackable, :validatable,
+         :jwt_authenticatable,
+         jwt_revocation_strategy: JWTBlacklist
```
brycemcd: Neat! I didn't know Devise could do this out of the box.

config/initializers/session_store.rb
```diff
@@ -0,0 +1 @@
+Rails.application.config.session_store :disabled
```
brycemcd: This might be a stupid question, but why disable sessions?

frontend/actions/userActions.js
```diff
@@ -0,0 +1,76 @@
...
+
+export function registerUser(params, successFn) {
+  return function registerUserThunk(dispatch) {
+    $.ajax({
+      url: '/api/signup',
+      type: 'POST',
+      data: params,
+      success: (data, textStatus, jqXHR) => {
+        console.log("successful registration")
+        console.log(jqXHR.getResponseHeader('Authorization')
```
brycemcd: You'll want to clean up these logs before going to prod. If my brain compiler is working (bit IF) you're missing a closing paren here.

frontend/components/LoginForm.jsx
```diff
@@ -0,0 +1,107 @@
...
+
+  validateRequiredFields() {
+    return this.state.email && this.state.password;
+  }
+
+  render() {
+    if (this.props.currentUser) {
+      return (
+        <div>
+          You already did this, you silly goose. go <a href="/success" onClick={push}>here</a>
```
brycemcd: <3

frontend/components/SuccessStub.jsx
```diff
@@ -0,0 +1,20 @@
...
+      <a href="/login" onClick={push}>Login</a><br/>
+    </div>
+  );
+};
+
+const mapStateToProps = state => ({
+  currentUser: state.users.currentUser,
+});
+
+export default connect(mapStateToProps)(SuccessStub);
```
brycemcd: You can have multiple argument groups in javascript now? Mind. Blown.

app/assets/stylesheets/success.css
```diff
@@ -0,0 +1,5 @@
+marquee {
+  font-size: 24px;
+  color: pink;
+  border: dotted green 3px;
+}
```
brycemcd: Does this need its own entire stylesheet? It's more intuitive, at least in my mind, to create a stylesheet like `login_and_registration` and dump all the styles needed for that function in one file.

Gemfile
```diff
@@ -9,6 +9,7 @@ end
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 gem 'rails', '~> 5.1.4'
+gem 'therubyracer', '~> 0.12.3'
```
e91096: 🏎 this sounds cool! what does it do?

config/initializers/devise.rb
```diff
@@ -7,6 +7,17 @@
...
   config.secret_key = ENV['DEVISE_SECRET_KEY']
+  config.jwt do |jwt|
+    jwt.secret = ENV['DEVISE_JWT_SECRET_KEY']
+    #jwt.dispatch_requests = [
+      ## ['POST', %r{^/api/signup$}],
+      ## ['POST', %r{^/api/login$}] TODO ???? does this even mean anything:???
+    #]
+    #jwt.revocation_requests = [
+      #['DELETE', %r{^/logout$}]
+    #]
```
e91096: should we keep this stuff around?

docs/authentication_and_registration.md
```diff
@@ -0,0 +1,107 @@
+# Primitive API documentation
```
e91096: awesome

Gemfile
brycemcd: This was necessary in order to get the rails server to start up. It's a server side javascript runtime. It's default IIRC and I imagine we can remove it if we further separate the FE from the BE.

config/initializers/devise.rb
brycemcd: :man_shrugging: I'm not sure. The backend doens't need it, but I'm not sure if the FE needs it.

PR #15, opened 2018-07-23
changes rails default to more appropriate license for PL

brycemcd:
The MIT license is too permissive IMHO. It allows for other organizations to profit and distribute their own version of our software and change the license. I don't have a problem with other organizations profiting from our software but I think we should require any such organization to make their version open source.

This license sets up an agreement that other organizations would be able to use our software but they would be required to attribute the original source, maintain the same license and share source code. The language appears to support our mission better.

More information about the new license here: https://choosealicense.com/licenses/agpl-3.0/


brycemcd:
Thea and I talked on the phone about this and the various implications. Quoth Thea:

> LGTM :+1:
> T. Briggs, c. 2018

I'm merging it!

PR #17, opened 2018-08-05
Submit votes

chrisruenes1:
Well, there is definitely a lesson here. Pull requests shouldn't be this big. In addition to the submit votes functionality, this PR includes a bunch of refactors and layout functionality.

 If this is unmanageable, I could find a way to split it into several pull requests. In the meantime, here is a high-level summary of the different features and refactors in the PR:

**Submit Votes**

This is the basic feature. I added a Campaign Show page that has an entry for each bill summary in a campaign. A user should be able to go to this page, and vote one of three ways on each bill: yes, no or abstain. (On the backend these options are converted into for, against and abstain, so as to integrate with the naming on the side of the sausage app. More on that later). For now, admins are able to change which campaign a bill summary belongs to, in the bill summary edit page. A user can get to the campaign show by going to the campaigns index, and clicking on the show button. We should audit the navigation and come up with a workflow, but I figure this works for now.

**Delete Dead Code**

There were a bunch of duplicate frontend files, named `bill` instead of `bill summary`. I deleted these files.

**Introduce Presenter Pattern**

The campaign controller now returns a set of nested presenters instead of raw json/jbuilder. This is a pattern I've used before and like a lot. Each presenter file inherits from the base_presenter, and aliases an `object` as the name of the class (ie `alias vote object`). Then the file just uses pure ruby to define exactly what the returned JSON should look like. Nested presenters are easy in this pattern.

**Introduce Service Object Pattern**

I took advantage of one of the requirements of this card, that we had to convert the name of the response between the frontend and backend, to introduce a services folder. In case the reviewer is unfamiliar, service objects are a great way to abstract non-boilerplate logic out of the controller. Generally, I like to move any non-generic functionality from controllers into service objects.

**Add stylesheets**

Including a reset and a clearfix.

**Break out the store into its own file**

This is desirable from a single responsibility principal perspective, and it also allows us to import the store and call `store.getState()` in multiple files

**Add a fetchIfEmpty abstract action**

This is one of those instances of metaprogramming that seems like it could either become very useful or very bloated/esoteric. We should keep an eye on whether the problem is consistent enough across files that it makes sense to abstract. It's kind of nice for now, though.

**Refactor the reducer patterns**

This is a big part of the MR. It took me a while to get my head around the suggestions in Dan Abramov's Building Idiomatic Redux Applications lectures, but I think I've got a pretty good handle on them now and worked to bring our reducer shape in line with those patterns. The key principal is to treat the store like a database; so, for each resource, there should be a `byId` reducer, that just stores each resource under the key of its id. Anything else should fit the pattern `idsBy<Resource>Id`, ie `idsByCampaignId`. Selectors, which are exported from the index of reducers per resource, handle mapping these sets of ids to the objects they point to. I definitely think it will keep our reducers clean and scalable to keep using these patterns as we go, so if there are questions about how they work, please let me know!

**Add a Navbar and Layout component**

This is pretty self-explanatory. It brings us much closer to having an actually navigable app! There is a new constants file on the frontend that has a NAV_ITEMS constant defined. To add a new navitem to the bar, just define its name and url. 

**Add a Select box to the bill summary form**

So that a bill summary can be assigned to a campaign on create/edit.

**Add a bunch of new helpers**

A side note: this file is becoming unwieldy. We should decompose it asap. 

**TO TEST**

Create a campaign or two, and then create a bunch of bill summaries for those campaigns. Make sure you are able to submit votes for each bill summary. Then, just try navigating around and trying out the CRUD functionality, attempting to break everything as you go. Make sure that database state persists across refresh of all pages, after back and forward in the browser, and after form submission. Try moving around between different forms and making sure that an update on one effects the data on another (ie, if I move a bill_summary in the edit to a different campaign, it should appear under that new campaign). 

**IN CONCLUSION**

I'm sorry about all this. I'll stand by most of these changes, but the unbroken refactor fervor definitely got a little out of hand. Again, if this is too much, I'm happy to try to decompose these changes to separate PRs. Otherwise, good luck, and let me know what comes up!


brycemcd:
Wow, this is a big honkin' PR :)

I see at least three reviewable things here:

1. Backend
2. Frontend
3. Presenters

I'm not sure what effect, if any, the changes in the FE code will have but I'm pretty sure that the presenter PR will influence #14. Having the backend (API) confirmed will help unblock some work we'd like to get done in the immediate future.

PR #20, opened 2018-08-07
Add google analytics

denebtech:


docker/env
```diff
@@ -0,0 +1,2 @@
+# this is the postgres user's (admin) password
+POSTGRES_PASSWORD=mysecretpassword
```
brycemcd: This file should be part of the `.gitignore`. We don't want this file in the repo

app/views/layouts/application.html.erb
```diff
@@ -7,6 +7,21 @@
...
     <link rel="stylesheet" href="https://unpkg.com/react-select/dist/react-select.css">
+
+     <!-- Global site tag (gtag.js) - Google Analytics -->
+    <script async src= <%= "https://www.googletagmanager.com/gtag/js?id=" + ENV["google_analytics_key"]  %>>
+
+    <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag(){dataLayer.push(arguments);}
+      gtag('js', new Date());
+      gtag('config', <%= ENV["google_analytics_key"] %>);
```
brycemcd: Does this get rendered as a string (i.e. "foo") or as a variable (i.e. `foo`)?

app/views/layouts/application.html.erb
```diff
@@ -7,6 +7,21 @@
     <%= stylesheet_link_tag    'application', media: 'all', 'data-turbolinks-track': 'reload' %>
     <%= javascript_include_tag 'application' %>
     <link rel="stylesheet" href="https://unpkg.com/react-select/dist/react-select.css">
+
+     <!-- Global site tag (gtag.js) - Google Analytics -->
+    <script async src= <%= "https://www.googletagmanager.com/gtag/js?id=" + ENV["google_analytics_key"]  %>>
```
brycemcd: Why not move the string part out into it's own part and use the ERB wrapper around the ruby part? i.e. `src="googletagmanager.com/gtag/js?id=<%= ENV['google_analytics_key']%>"`

docker/env
andrezafeu: @brycemcd I think we should also add logfile to .gitignore

denebtech: I just added `!/log/.keep` to .gitignore.

brycemcd: please remove this file from the PR

brycemcd:
Please remove that file and then :shipit: (merge to master). LGTM

PR #21, opened 2018-08-08
Submit votes frontend

chrisruenes1:
PLEASE MERGE IN SUBMIT VOTES BACKEND (#23) BEFORE REVIEWING THIS BRANCH

This is the most substantial pull request of the submit votes trilogy. 

It includes all the boilerplate necessary for a compound form in the campaign show page, comprised of a VoteForm for each bill_summary associated with a campaign.

In addition, it includes some major refactors of the frontend. The biggest changes are to the reducer structure. I followed some suggestions from Dan Abromov's "Idiomatic Redux" videos, so that each reducer folder has a `byId` reducer. This reducer acts like a datastore, and others in the same folder just associate a single foreign key id to an array of ids. I'm happy to do a walkthrough if the reviewer is unfamiliar with this pattern, as it took me a while to get a handle on.

I also added structural components, ie a Layout component, and navbar. (The constants.js file was necessary for my implementation of the navbar, but I see it as potentially useful moving forward as well).

I added a stylesheets folder as well. I included a reset and a clearfix.

I added a `fetchIfEmpty` global action. I talk more about this in the commit history, but it should take care of making sure that all navigation (including page refresh) leaves us a hydrated store for now.

I made notes in the commit history, so feel free to check it out with other questions

Once you have, please just make sure that all navigation and CRUD functionality works as expected throughout the app. Try as hard as you can to find edge cases that might break the flow. Since these changes touch almost all of the frontend, naive clicking around the app may be a good way to discover bugs.


frontend/routes.jsx
```diff
@@ -15,7 +16,8 @@ const routes = [
   { path: /\/billSummaries\/(\d+)$/, action: () => <BillSummaryForm /> },
   { path: /\/campaigns$/, action: () => <CampaignsIndex /> },
   { path: /\/campaigns\/new/, action: () => <CampaignForm /> },
-  { path: /\/campaigns\/(\d)$/, action: () => <CampaignForm /> },
+  { path: /^\/campaigns\/(\d+)$/, action: () => <CampaignShow /> },
```
e91096: should we actually add `^` to all our route regexes?

app/controllers/api/votes_controller.rb
```diff
@@ -4,7 +4,7 @@ def create
     current_user_id = 1
     vote = Vote.new(permitted_params.merge(user_id: current_user_id))
     if vote.save
-      render json: vote
+      render json: vote.as_json
```
e91096: 🤔🤔🤔

frontend/actions/campaignActions.js
```diff
@@ -1,4 +1,5 @@
 import * as $ from 'jquery';
+import { backendToFrontendParams } from './submitVotes';
```
e91096: should we specify that these are `voteParams`? in case we ever have/need another action with this name?

frontend/constants.js
```diff
@@ -0,0 +1,12 @@
+export const NAV_ITEMS = [
```
e91096: this is great

frontend/helpers/index.js
```diff
@@ -24,6 +24,21 @@ export function formatOptionsForSelect(arr, opts = { valueKey: 'id', labelKey: '
...
+// TODO I feel like we're going to want this even now nothing is pointing to it anymore
+// export function arrayToSelectOptions(arr) {
+//   return arr.map(item => ({
+//     label: item,
+//     value: item,
+//   }));
+// }
+
+export function removeItem(list, itemToRemove) {
+  // TODO check whether there is a simple underscore way to do this, once internet gets invented
```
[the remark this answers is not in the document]
e91096: `_.without`?

frontend/routes.jsx
chrisruenes1: oooooooooh yeah we should!

frontend/actions/index.js
```diff
@@ -0,0 +1,53 @@
...
+            ...objKeysToSnakeCase(params),
+          },
+        } : {};
+        $.ajax({
+          url,
+          data: requestData,
+          success: data => dispatch(receiveResources(snakeCaseResource, data)),
+          error: (data) => {
+            // TODO build in error handling from backend
+            console.log(data);
```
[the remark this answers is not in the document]
chrisruenes1: Hahaha yes we should certainly get to this change sooner rather than later. @e91096 you were working on an MR for error handling, if I remember correctly, right? Did that include anything on the frontend? @brycemcd I haven't encountered mixpanel before, would you mind elaborating a bit?

frontend/actions/submitVotes.js
```diff
@@ -0,0 +1,65 @@
...
+      $.ajax({
+        url,
+        type: updateOrCreate,
+        dataType: 'json',
+        contentType: 'application/json',
+        data: JSON.stringify(params),
+        success: data => dispatch(receiveVote(data)),
+        error: (data) => {
+          // TODO build in error handling from backend
+          console.log(data);
```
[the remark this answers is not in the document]
chrisruenes1: 👍

frontend/helpers/index.js
```diff
@@ -24,6 +24,21 @@ export function formatOptionsForSelect(arr, opts = { valueKey: 'id', labelKey: '
   return arr.map(item => formatItemForSelect(item, opts));
 }
 
+// TODO I feel like we're going to want this even now nothing is pointing to it anymore
```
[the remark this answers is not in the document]
chrisruenes1: Hahaha I'm gonna get rid of this one, it wasn't that hard to implement and the helpers file is quite bloated already. Speaking of which, I'd like to decompose it in another PR!

frontend/helpers/index.js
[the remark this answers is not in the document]
chrisruenes1: hahaha yes definitely _.without 😂

app/controllers/api/votes_controller.rb
chrisruenes1: yes well

this goes back to a longer discussion that Bryce and I had on Slack (@brycemcd would have been a good call to move it to the PR as you suggested). In overriding the inherited `as_json`, and calling it implicitly through `render json`, you need to specify an argument in the method that isn't actually used. Bryce pointed out that this wasn't ideal, and we changed it in the backend PR, but this controller got through the cracks. I'm open to continuing to discuss, but figured we could wait til we get to the presenter PR to do so, since it falls into that anyway.

frontend/actions/campaignActions.js
chrisruenes1: Yes good call!

frontend/constants.js
chrisruenes1: 😄

chrisruenes1:
@e91096 @brycemcd I implemented the feedback above, except for: 

1. Error handling on the frontend
2. Testing

Would everyone be comfortable with making these issues separate cards for now, after some discussion?

frontend/actions/index.js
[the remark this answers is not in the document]
brycemcd: IIRC, we closed Esther's error formatting PR because we more or less solved for it in #14 .

This is an early app without any traffic at the moment so I'm fine deferring this to another card. In the future though, when even one person is using the app, we'll want to make sure to communicate errors to the user && log the error somehow.

Mixpanel is an application for user event tracking. We can do things like increment a counter when a specific user encounters an error. We can then get a dashboard of those errors and create & prioritize them.

brycemcd:
I'm comfortable separating error handling and testing into separate cards given the dependencies on this card we've built into other PRs.

PR #22, opened 2018-08-08
Replace campaign json with presenters

chrisruenes1:
PLEASE REVIEW SUBMIT VOTES BACKEND (#23) BEFORE REVIEWING THIS BRANCH

This is a (refreshingly) small one. Introduce the presenter pattern, which standardizes API responses. Presenters use plain ruby, which is nice if you've ever had to mess around with jbuilder or the like before. 

In a presenter, we overwrite the `as_json` method so that we can easily `render json: <SomePresenter>.new(<some_model>)` in the controllers.


brycemcd:
I think you're on the right track with the idea of presenters but I think we could improve the implementation a bit.

Given that this is an API-only backend, it's my personal opinion that the view layer in the MVC framework should construct and represent the json responses. Consider that if the server were rendering html templates, the view layer would be responsible for compiling the templates and sending them down the wire. As an API there's no need for html templates, obviously, but the convention can be reused properly with json templates.

Looks like [jbuilder](https://github.com/rails/jbuilder#jbuilder) is the conventional way to render templates in json in rails now.

@e91096 and I put some thought into returning [a consistent JSON object](https://github.com/PocketLobby/constituents/pull/14/files#diff-55c5b7aecfb519d0e4880eaf2788eb6eR28) to the FE in our work with user auth + auth. The idea is that regardless of the request, the response is familiar. There's a lot of room for improvement to be sure, but it seems we've embarked on this effort at approximately the same time :)

I don't think we need a presenters directory. IMHO, we can use the views as I said before. Another thing to think about is how we can allow the domain objects (votes, bill summaries etc) to render themselves in json in a consistent way; we don't want to have to keep writing the same hashes over and over and then convert them to json (that's more of a criticism of the user auth PR rather than this PR).

e91096:
what does the presenter abstraction give us over making it a helper method in the controller for instance?

chrisruenes1:
Hey @brycemcd and @e91096! Sorry I've taken a while to respond to these comments, I had a crazy last week and am just returning from visiting my family in PA.

So, @brycemcd I lifted this pattern from a coworker of mine who converted all of his team's jbuilder view files into presenters. My main arguments against using jbuilder:

1. It's really slow. This obviously isn't a problem for us now, but even on my app at work, which has only ~100 users, we found that transitioning from jbuilder to pure Ruby noticeably improved response time
2. It relies on a whitespace based DSL that I find much less intuitive than Ruby. If I'm being honest, my dislike of this DSL is probably my biggest motivation not to use jbuilder, and I think it was too for my coworker that initiated the  transition to using presenters. Admittedly, I generally dislike most DSLs (I even have some light beef with AR), and so my views on this specific facet of jbuilder might be a bit outside the mainstream. So, if you don't feel this way about them, I'm happy to concede on this point. I just don't like having to look up the syntax every time I want to write a view, ya know? 

@e91096 It definitely would work to put a helper method in the controller (and, I like your addition to the application controller!).  I think the benefit of using presenters is that they basically serve the same function as views would, but instead of using jbuilder to enable the views to render json, they just use ruby to preparer a hash, which can be rendered automatically as json.

So it sounds like the three options we'e considering for rendering json are:

1. Use views + jbuilder
2. Use a ruby-based presenter pattern with a `presenters` folder that serves the role that the `views` folder normally would
3. Use a helper method that each controller inherits from `application_controller`, and can overwrite as necessary. 

I lean slightly towards 2, but I think that:

- 1 is good in that Rails is a framework that is based in convention, and 1 is definitely the Rails solution to this problem
- 3 is good in that it would likely cut down on a lot of boilerplate

Let me know what you both think!

brycemcd:
That's good stuff @chrisruenes1!

I wasn't aware of the performance implications of JBuilder. That's shocking. But, I think it would be a pre-optimization to toss out the rails convention for a problem we don't have :)

FWIW, according to the docs I've read, it doesn't depend on a whitespace DSL. It's a DSL for sure, but it seems like you can space however the heck you want. I haven't actually tried it in my editor.

The questions I would lean on to guide this decision:

* Do we have a solid reason for not going with the rails convention?

> Going _with_ the convention includes a community of a lot of nerds that have generated a lot of documentation and bug reports. Adding new nerds to our project is easier. Tossing the convention may be warranted but we should articulate the tradeoffs we're making.

* Are we using the right level of abstraction?

> We're in a MVC framework. Whatever we go with, lets make sure that we appreciate the different aspects of this architecture and insert our solution in the right component.

* Are we able to represent a domain object once and include that in many responses? (this is basically the dry principle.)

> Lots of boilerplate is not fun and leads to stupid human errors. Idioms like `render json: object` makes the controller stupid easy to read. More complicated endpoints where we might do something like `render json: { votes: current_user.votes, bills: current_user.bills }` provides a thought experiment where we can see that we'd want to reference `votes` in more than one place but define the domain object of `votes` in one place so the API can evolve without having to remember where each `vote` was used.

chrisruenes1:
@brycemcd So I prefer the presenter pattern in the abstract, but I think that using `jbuilder` until it causes problems (hopefully, since that will mean out network traffic is substantial enough for that to happen :) ) makes sense from a Go With The Convention For Onboarding Reasons perspective. So I'm going to close this PR. Do you want me to make a card for the backend team to switch over to jbuilder at some point?

PR #23, opened 2018-08-08
Submit votes backend

chrisruenes1:
Create the API for votes!

Most of this is pretty boilerplate, but there are a few things to note:

1. Votes belong to both a `bill_summary` and a `campaign`. Since `bill_summary`s also belong to `campaigns`, this denormalizes the relationship. It may be a little early for that, but it was a solution for how to allow campaigns to `accept_nested_attributes` for `votes`, without having to add another purely structural model.

2. I introduced the service object pattern in this PR. I would like for any non-boilerplate logic in the controllers to be abstracted to service objects. Here, I created a service object to translate between backend and frontend naming for the vote responses.


app/controllers/api/campaigns_controller.rb
```diff
@@ -35,16 +45,26 @@ def destroy
...
+    {
+      id: campaign.id,
+      name: campaign.name,
+      votes: campaign.votes.map do |vote|
+        id: vote.id,
+        response: VoteResponseResolver.new(vote.response, :frontend).resolve,
+        bill_summary_id: vote.bill_summary_id,
+        campaign_id: campaign.id
+      end
+    }
```
brycemcd: Shouldn't this be in the model (where `Campaign` is defined)? IIRC, if you override the `to_json` method in the model, you can simply call `render json: campaign` in the controllers.

app/models/vote.rb
```diff
@@ -0,0 +1,10 @@
+class Vote < ApplicationRecord
+  belongs_to :bill_summary
+  belongs_to :campaign
```
brycemcd: I don't think this is quite right. I saw that you mentioned this in your PR summary, but it feels off to me that a vote should depend on a campaign. See note in campaign.rb.

app/models/campaign.rb
```diff
@@ -1,4 +1,7 @@
 class Campaign < ApplicationRecord
   validates :name, presence: true, uniqueness: true
   has_many :bill_summaries, dependent: :destroy
+  has_many :votes, dependent: :destroy
```
brycemcd: I think you want this to be `has_many :votes, through: :bill_summaries`. Doing so will release the dependency to collect a campaign id for each vote. TBH, I'm pretty sure we can consider campaigns and votes completely separate entities; I don't think we have a need to get to votes through campaigns.

config/routes.rb
```diff
@@ -3,7 +3,9 @@
   devise_for :users
   namespace :api do
     resources :bill_summaries, only: [:index, :show, :create, :update, :destroy]
-    resources :campaigns, only: [:index, :show, :create, :update, :destroy]
+    resources :campaigns, only: [:index, :show, :create, :update, :destroy] do
+      resources :votes, only: [:index, :show, :create, :update, :destroy]
```
brycemcd: Again, I don't think we need to make votes dependent on campaigns.

app/controllers/api/campaigns_controller.rb
```diff
@@ -35,16 +45,26 @@ def destroy
...
   def campaign_json(campaign)
-    #TODO replace this with prersenter pattern
-    { id: campaign.id, name: campaign.name, bill_summary_count: campaign.bill_summaries.count }
+    # TODO take this out once the presenter PR is merged
+    {
+      id: campaign.id,
+      name: campaign.name,
+      votes: campaign.votes.map do |vote|
+        id: vote.id,
+        response: VoteResponseResolver.new(vote.response, :frontend).resolve,
```
brycemcd: This + the formatter + the resolver seems like an unnecessary bit of indirection to me. If my brain compiler is working right, you're returning a representation of the campaign with votes. The resolver is a simple mapping class from one domain language (i.e. "yes") to another (i.e. "for") but I think the API should be defined simply in terms of the enum ("for", "against", "abstain") we want in the data while the FE can present those terms any way it wants.

Further, I don't think CRUD'ing a campaign should be CRUD'ing or representing votes as well. They're independent entities. This controller should be concerned with only CRUD'ing a campaign. Another controller should CRUD the _relationship_ between the campaign and its bill summaries and yet another controller should be concerned with creating the votes.

app/controllers/api/campaigns_controller.rb
```diff
@@ -35,16 +45,26 @@ def destroy
     campaign = Campaign.find(params[:id])
     campaign.destroy
     render json: campaign_json(campaign)
```
brycemcd: This should probably just return a 204 shouldn't it?

app/controllers/api/votes_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def destroy
+    vote = Vote.find(params[:id])
+    if vote.destroy
+      render body: nil, status: 204
+    else
+      render json: vote.errors.full_messages
+    end
+  end
```
brycemcd: Do we want `show` and `destroy` actions?

app/models/campaign.rb
```diff
@@ -1,4 +1,13 @@
...
   has_many :bill_summaries, dependent: :destroy
+  has_many :votes, through: :bill_summaries
+
+  def as_json(*)
+    # TODO take this out once the presenter PR is merged
+    {
+      id: id,
+      name: name,
+    }
+  end
```
brycemcd: Unless there's a compelling reason, I don't see the need for an anonymous splat parameter argument (`as_json(*)`) It looks like that could be removed and the method could be reduced to:

```ruby
{id: self.id,
name: self.name,
}
```

app/models/vote.rb
```diff
@@ -0,0 +1,18 @@
...
+  scope :for_campaign, ->(campaign_id, user_id) { joins(:bill_summary).where({ bill_summaries: { campaign_id: campaign_id }}, user_id: user_id) }
+
+  def as_json(*)
+    #TODO remove once presenters merged in
+    {
+      id: id,
+      response: response,
+      bill_summary_id: bill_summary_id
+    }
+  end
```
brycemcd: Same comment here as in `Campaign#as_json`

app/controllers/api/votes_controller.rb
andrezafeu: Done

PR #25, opened 2018-08-09
Add redirect after campaign

andrezafeu:
Implemented:
- Redirect after a campaign is complete;
- Read redirect URL from an environment variable.

Original story was: As an alpha tester who has just completed a campaign, I should be redirected to a Google Form where I will be asked for NPS and other questions that may come up.

We will probably have to come back to this after the voting code is complete.


andrezafeu:
It looks like I messed up the rebase.
I've pushed a new branch with the changes, so I wouldn't have to force a push. PR: #31

PR #28, opened 2018-08-15
Add style to campaign pages

andrezafeu:
- Configure css stylesheets for react (I followed the configuration from PR#21)
- Add style to campaign form.


frontend/components/CampaignForm.jsx
```diff
@@ -52,20 +55,35 @@ class CampaignForm extends Component {
...
-            id="campaign-name"
-            label="Name"
-            handleChange={this.handleCampaignNameChange}
-            type="input"
-            validateInput={() => []}
-            value={this.state.name}
-          />
+      <div className="containerBody">
+        <div className="backgroundBanner">
+        </div>
```
e91096: this can be a self-closing tag!
`<div className="backgroundBanner" />`

frontend/components/CampaignForm.jsx
```diff
@@ -52,20 +55,35 @@ class CampaignForm extends Component {
...
-        <div>
-          <Input
-            id="campaign-name"
-            label="Name"
-            handleChange={this.handleCampaignNameChange}
-            type="input"
-            validateInput={() => []}
-            value={this.state.name}
-          />
+      <div className="containerBody">
```
e91096: i prefer kabob case for css names. thoughts/opinions @brycemcd @chrisruenes1?

frontend/components/CampaignForm.jsx
```diff
@@ -52,20 +55,35 @@ class CampaignForm extends Component {
...
+        <div className="backgroundBanner">
+        </div>
+        <div className="formInfo">
+          <div className="formTitle">
+            <h1><Input
```
e91096: can we put this `<Input>` on its own line?
and also the `</h1>` below

frontend/stylesheets/forms.css
```diff
@@ -0,0 +1,91 @@
+h1 {
```
e91096: i think the name of this file can be something more specific to the campaign page--at least until we need a more generally-usable form style sheet.

e91096: also, i'm not sure if these header styles should apply globally. can we namespace them to the specific form we're styling here?

frontend/components/CampaignForm.jsx
```diff
@@ -52,20 +55,35 @@ class CampaignForm extends Component {
...
+              id="campaign-name"
+              label="Name"
+              handleChange={this.handleCampaignNameChange}
+              type="input"
+              validateInput={() => []}
+              value={this.state.name}
+            /></h1>
+          </div>
+          <div className="formInstructions">
+            <h6>* Required</h6>
```
e91096: i don't know if `* Required` should use a header tag

frontend/components/CampaignForm.jsx
e91096: actually, should there be an input field here?

frontend/stylesheets/forms.css
```diff
@@ -0,0 +1,91 @@
...
+}
+
+p {
+  margin-top: 1em;
+  margin-bottom: 1em;
+}
+
+.containerBody {
+  background-color: #DDD;
+  font-family: "Roboto, RobotoDraft, Helvetica, Arial, sans-serif";
```
e91096: any idea why i might still be seeing serif font? 
screen shot 2018-08-20 at 3 14 13 pm

frontend/components/CampaignForm.jsx
andrezafeu: ok, I will change to paragraph

frontend/stylesheets/forms.css
andrezafeu: They wouldn't be applied globally, only to the pages that are linking to this css sheet (which will probably be used only on form style pages). Do you still think I should namespace them?

e91096: OH OH i understand! sorry i didn't realize what was happening! i think this is fine then!

frontend/components/CampaignForm.jsx
brycemcd: It's your FE. Let's establish a standard, communicate that through our contribution guidelines (or some other style guide) + Slack and then enforce it going forward.

frontend/stylesheets/forms.css
brycemcd: Check the "computed" tab in the inspector. What font is rendering? The code looks right to me but it could be that another style rule is taking precedence.

frontend/components/CampaignForm.jsx
```diff
@@ -1,10 +1,13 @@
 import React, { Component } from 'react';
```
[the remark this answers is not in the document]
brycemcd: We'll probably want it to style all the campaign related pages. That will come in a future improvement. For now, I just wanted to get some of the rules and code established while we had Andreza freed up.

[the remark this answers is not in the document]
e93285: ah, fair enough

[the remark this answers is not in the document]
e91096: fair enough

frontend/components/CampaignForm.jsx
```diff
@@ -52,20 +55,36 @@ class CampaignForm extends Component {
...
+            </h1>
+          </div>
+          <div className="form-instructions">
+            <p>* Required</p>
+          </div>
+          <div className="form-classification">
+            <h3>Category Header</h3>
+          </div>
+          <div className="form-description">
+            <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum. <a href="http://www.site.com">http://www.site.com</a></p>
```
e91096: let's take this and `Category Header` out of this page and maybe add these styles to the now-merged campaign show page to make sure they work well with other components!

PR #29, opened 2018-08-19
Support case-insensitive bill numbers.

chrisruenes1:
The regex was testing only for lower case letters. Rather
than complicating the regex and de-standardizing the way
we represent bill numbers, I added a formatter in
the controller that converts everything to downcase


app/controllers/api/bill_summaries_controller.rb
```diff
@@ -53,4 +55,9 @@ def destroy
   def permitted_params
     params.require(:billSummary).permit(:campaign_id, :summary, :bill_number, :full_bill_url)
   end
+
+  def format_params(params)
+    params[:bill_number] && params[:bill_number] = params[:bill_number].downcase
```
e91096: nice!

chrisruenes1:
@e91096 I moved the formatting conversion over to the FE!

PR #32, opened 2018-08-23
Update docker README

andrezafeu:
Update docker README


docker/README.md
```diff
@@ -24,6 +26,14 @@ Running backend tests is easy:
 
 `docker-compose run --rm constituent_rails rspec`
 
+## Migrations
+
+`docker-compose run --rm constituent_rails` + migration command
```
e91096: this works for me!
just curious: what's the different between `docker-compose run --rm constituent_rails` and `docker-compose exec constituent_rails`?

docker/README.md
```diff
@@ -24,6 +26,14 @@ Running backend tests is easy:
...
+
+Example:
+
+`docker-compose run --rm constituent_rails rake db:migrate`
```
e91096: can we use the whole `bundle exec rails db:migrate` as the migration command?

docker/README.md
andrezafeu: `exec` runs a command in an existing container, while `run` runs it on a new one

e91096:
👍

PR #33, opened 2018-08-28
Add status to campaign

andrezafeu:
Campaign is Live (https://trello.com/c/s3eW3XUG/17-campaign-is-live)

Adds a boolean field to campaigns to indicate if a campaign is active or not.  Only one campaign can be active at a given time.


frontend/components/CampaignShow.jsx
```diff
@@ -44,6 +44,15 @@ class CampaignShow extends Component {
     return (
       <div>
         <div>{this.props.campaign.name}</div>
+        <div>
+          <label for="campaign-status">Active?</label>
```
e91096: i think you might want `htmlFor` here!

frontend/components/CampaignShow.jsx
```diff
@@ -44,6 +44,15 @@ class CampaignShow extends Component {
...
+          <input
+            id="campaign-status"
+            type="checkbox"
+            value={this.props.campaign.active}
+            checked={this.props.campaign.active}
```
e91096: should the user be able to manipulate this checkbox?

frontend/components/CampaignForm.jsx
```diff
@@ -62,6 +74,15 @@ class CampaignForm extends Component {
             validateInput={() => []}
             value={this.state.name}
           />
+          <label for="campaign-status">Active?</label>
+          <input
+            id="campaign-status"
+            onClick={this.toggleCampaignStatus}
+            type="checkbox"
+            validateInput={() => []}
```
e91096: you don't need this `validateInput` prop here

e91096:
i'm seeing some weird behavior--when i set an inactive campaign to active, and it takes me back to the campaign index page, i still see the previously active campaign at the top of the list with active true.
aug-28-2018 01-50-16

frontend/components/CampaignShow.jsx
andrezafeu: I think for now yes. There's another PR that adds cancancan to manage user abilities. Perhaps we can restrict it after that gets merged?

e91096: ah okay that's what i thought--i can't seem to edit it. clicking it has no affect on the value of the checkbox

e91096: note: this is from the campaign **show** page. it works just fine from the edit page!

frontend/components/CampaignsIndex.jsx
```diff
@@ -15,18 +15,33 @@ class CampaignsIndex extends Component {
...
+        } else {
+          return -1;
+        }
+      });
     }
+
     return this.props.campaigns.map((campaign) => {
       const cells = [
         campaign.name,
+        campaign.active.toString(),
```
e91096: i'm getting the following js error here when i update a campaign's status and submit
`TypeError: Cannot read property 'toString' of undefined`

frontend/components/CampaignsIndex.jsx
```diff
@@ -71,9 +87,10 @@ const mapDispatchToProps = dispatch => ({
 
 CampaignsIndex.propTypes = {
   campaigns: PropTypes.arrayOf(PropTypes.shape({
-    bill_summay_count: PropTypes.number.isRequired,
+    bill_summary_count: PropTypes.number.isRequired,
```
e91096: ❤️

e91096:
i still can't seem to manipulate the checkbox from the show page. if that's fine, then i'm okay to plus 1! 
sep-08-2018 13-53-00

e91096:
@andrezafeu i'm also seeing a crazy amount of network requests to `/api/campaigns` when i'm on the campaigns index page...

andrezafeu:
@e91096 I think the user shouldn't be able to manipulate the status of a campaign. Only the admin, right? To keep the app consistent, I think neither the user or admin be able to edit the status in the show page, only through the update link.  That being said, I do agree that having a checkbox in the show page is confusing. But perhaps this is more of a design issue? Also, should the user be able to access campaigns that are not active? If not, we can just hide that field in the show page.
PS: I will take a look at the network requests

e91096:
yeah, the code isn't written in a way that would make me think it's an editable checkbox. just wanted to make sure after this convo!: https://github.com/PocketLobby/constituents/pull/33#discussion_r213301348. 
i think you're right that we should probably think about the design ¯\_(ツ)_/¯

otherwise this looks great! do you mind rebasing so i can test with all the latest code?

andrezafeu:
@e91096 I rebased the branch and added a question in Trello about the campaign status in the show page!

PR #35, opened 2018-08-30
Generate invite tokens to users

andrezafeu:
[Constituent User Invitations](https://trello.com/c/YKQSueR0/12-constituent-user-invitations)

## Dependent PRs
[Accept User Invitations](https://trello.com/c/bomO7FOx/13-accept-user-invitations)


db/migrate/20180829221929_add_invite_token_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddInviteTokenToUsers < ActiveRecord::Migration[5.1]
+  def change
+    add_column :users, :invite_token, :string, null: false
```
e91096: let's add a uniqueness db constraint, too

frontend/components/SuccessStub.jsx
```diff
@@ -7,6 +7,12 @@ import { logoutUser } from '../actions/userActions';
 const SuccessStub = (props) => {
   return (
     <div>
+      <h1>
+        Thanks for registering!
+      </h1>
+      <h2>
+        Your invite URL is {window.location.origin}/register/{props.currentUser.invite_token}
```
e91096: 2 questions: 1. this doesn't lead anywhere right now, right?
2. should we either redirect successful **log in** to a different page, OR conditionally show these registration-specific lines only when the user has been redirected here after successful registration?

andrezafeu: 1. Right
2. I'm not sure what Bryce had in mind. But, in my opinion, if the goal is to get a lot of users quickly, this should be shown in the login page too. The more often a user sees the registration link, more likely he/she is to invite other people.  I do think that when the design is ready, we can rethink this if the design is 'too cluttered' : )
Any thoughts @brycemcd ?

brycemcd: I think the feature is not fully spec'd :) The decisions you're making are great. We know that we want to allow users to share their invitation link and have others consume it and the reasoning that you're applying above is great.

This is software and it's malleable. Let's do what you think is best for now and when we get more detailed specs we can change the implementation as needed.

e91096: that's a really great point!
I do think `Thanks for registering!` should only appear if they're redirected after registration; it could be a confusing user experience otherwise

andrezafeu: Yeah, makes sense! I had forgotten about this 'Thanks for registering' thing. I created a new page for success after registration. I didn't create a partial cause the contents are not defined yet... right?

frontend/components/SuccessStub.jsx
```diff
@@ -7,9 +7,11 @@ import { logoutUser } from '../actions/userActions';
 const SuccessStub = (props) => {
   return (
     <div>
+      <h2>
+        Your invite URL is {window.location.origin}/register/{props.currentUser.invite_token}
+      </h2>
       <a href="/billSummaries" onClick={push}>Bill Summaries</a><br/>
       <a href="/campaigns" onClick={push}>Campaigns</a><br/>
-      <a href="/login" onClick={push}>Login</a><br/>
```
e91096: good call

andrezafeu:
Yeah, a single success page would probably be a better idea. This way we can play with alerts, that can be easily dismissed by users.

PR #36, opened 2018-08-30
Epic spec and CI fails

brycemcd:
[Slack Message](https://pocketlobby.slack.com/archives/C7Q12JY21/p1535594375000100)

## Prerequisite PRs

None

## Dependent PRs

None

## Notes for Reviewer

We should have a postmortem. Looks like this was a bad merge coupled with some unexpected behavior in CI.

There were some failing tests and an errant string in one of the test files which was causing an exception to be thrown.

I had assumed (because I set it up) that CI ran all the tests written in ruby prior to giving the :heavy_check_mark: . It turns out that the tests were stubbed in CI and never run. Sad. Face.

Since then, we've implemented docker-compose.

This PR:

1. fixes the failing tests
2. reconfigures CI to run the _ruby_ tests
3. reconfigures CI to be more like our docker environment
4. removes some authorization calls (to be re-addressed in our auth/auth goal)


frontend/index.jsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { Provider } from 'react-redux';
 import ReactDOM from 'react-dom';
 
-import Layout from './components/Layout';
+//import Layout from './components/Layout';
```
e91096: is this unused?

e91096: update: it does not! revert?

brycemcd: OOPS

brycemcd:
@e91096 I updated the index.jsx. Take a peek and lemme know what you think.

PR #37, opened 2018-09-01
Removes Figaro, adds env to docker

brycemcd:
See the README for more detail. The dependency for Figaro
no longer exists in the docker world. Modifying the app
such that environment dependencies are managed in our docker
files. Also adds a workflow for adding required environment
variables to an initializer so that the app will fail to start
if any required keys are missing.

This is an OPPREF

## Prerequisite PRs
No Prerequisites, but I built this on #36 to minimize the diff

## Dependent PRs
No strict dependencies, but I built this can easily follow #36 

## Notes for Reviewer

Boot the app without some dependent env vars and notice how it fails to boot. Sets a convention for having all env vars be capitalized strings.


brycemcd:
merged into master manually

PR #39, opened 2018-09-02
[wip] write votes to queue

e91096:
[Trello Ticket](https://trello.com/c/Hr2a3QQ2/27-write-votes-to-queue)

## TODO/TBD:
1. user token
2. write to queue on vote update?
3. is this an okay format for the message?
     ```
     {
        user_token: "string",
        bill_id: "string",
        campaign_id: int,
        vote: "string",
        timestamp: timestamp,
     }
     ```

## New ENV Vars
AWS_ACCESS_KEY_ID
AWS_QUEUE_URL
AWS_REGION
AWS_SECRET_ACCESS_KEY

## Notes for Reviewer
does it make sense to do this in the votes controller?
docs: https://docs.aws.amazon.com/sdk-for-ruby/v3/developer-guide/sqs-example-send-and-receive-messages.html


brycemcd:
@e91096 I'd prefer to keep to the format that was specified in the ticket. Sausage already processes messages like that format so there wouldn't be any dependent code changes in that app. Is that a big ask?

I don't think it makes sense to do this in the controller as it is not a request/response concern. Rather, it's closer to a persistence concern. It's probably best factored as its own service class that can be called periodically either on an `after_commit` hook or on a regular schedule with something like cron.

app/controllers/api/votes_controller.rb
```diff
@@ -14,7 +21,16 @@ def create
...
+
+    if vote.update(vote_params)
+      # VoteQueuer.write_to_queue(
+      #   vote.user.id, # TODO user_token
+      #   vote.bill_summary.bill_number,
+      #   vote.bill_summary.campaign.id, # TODO can votes have a campaign?
+      #   vote_params[:response],
+      #   Time.now,
+      # )
+
```
brycemcd: IMHO, you're thinking the right way. The duplication of the code suggests that you may be better off passing a message to the `VoteQueuer` class like this when a vote gets inserted/updated:

`VoteQueuer.write_message(vote.id)`

Then the `VoteQueuer` class can look up the vote and write it to the SQS queue.

app/models/vote.rb
```diff
@@ -2,6 +2,8 @@ class Vote < ApplicationRecord
   belongs_to :bill_summary
   belongs_to :user
 
+  # delegate :campaign, :to => :bill_summary, :allow_nil => true
+
```
brycemcd: errant comment?

app/models/vote_queuer.rb
```diff
@@ -0,0 +1,22 @@
+require 'aws-sdk-sqs'
+
+class VoteQueuer < ApplicationRecord
+  def self.write_to_queue(user_token, bill_id, campaign_id, vote, timestamp)
+    sqs = Aws::SQS::Client.new(
+      region: ENV['AWS_REGION'],
+      access_key_id: ENV['AWS_ACCESS_KEY_ID'],
+      secret_access_key: ENV['AWS_SECRET_ACCESS_KEY'],
+    )
```
brycemcd: Do you think this bit could be moved out into its own config method?

app/models/vote_queuer.rb
```diff
@@ -0,0 +1,22 @@
+require 'aws-sdk-sqs'
+
+class VoteQueuer < ApplicationRecord
+  def self.write_to_queue(user_token, bill_id, campaign_id, vote, timestamp)
```
brycemcd: Something to consider:
If you accept a vote_id as the only parameter, then you can look up that vote record in this method and have access to all its methods to properly construct a message. The benefit is that as the vote domain object evolves, the message that gets written to the queue may or may not evolve along with it. Either way, you reduce coupling the implementation of the method to the current construction of the vote model (i.e. this method signature doesn't need to change as the vote model changes).

config/initializers/required_env_vars.rb
```diff
@@ -4,6 +4,10 @@ class EnvironmentConfigurationError < StandardError
 end
 
 required_keys = [
+  "AWS_ACCESS_KEY_ID",
+  "AWS_QUEUE_URL",
+  "AWS_REGION",
+  "AWS_SECRET_ACCESS_KEY",
```
brycemcd: YAY - it works!

db/seeds.rb
```diff
@@ -47,3 +61,15 @@ def create_records(table_string)
...
+create_records('votes') do
+  votes = (1..5).map do |id|
+    Vote.create({
+      id: id,
+      bill_summary_id: BillSummary.find(id),
+      user_id: 1,
+      response: ["yes", "no", "abstain"].sample,
+    })
+  end
+end
```
brycemcd: nice. Well done

docs/PULL_REQUEST_TEMPLATE.md
```diff
@@ -9,4 +9,8 @@
 
 
 
+## New ENV Vars
```
brycemcd: Great idea

brycemcd:
LGTM :+1:

PR #40, opened 2018-09-03
Fixing CI ... again

brycemcd:
CI broke because of a step in the asset pipeline. I hadn't tested busting the cache. This PR fixes that.

## Prerequisite PRs
None

## Dependent PRs
None

## Notes for Reviewer
Try it with your code. It should work.


brycemcd:
That's more or less the approach I was taking with testing this. CI is passing, compiling assets _and_ actually running the ruby tests.

chrisruenes1:
Great then LGTM!

PR #41, opened 2018-09-04
Wip add user token to campaign rebased

andrezafeu:
[Campaign Landing Page](https://trello.com/c/YukBnzlk/25-campaign-landing-page)

## Notes for Reviewer
Comment out `before_action :authenticate_api_user!` on votes controller to test code


app/models/vote.rb
```diff
@@ -4,6 +4,7 @@ class Vote < ApplicationRecord
 
   validates :response, presence: true, inclusion: { in: %w(for against abstain) }
   validates :bill_summary_id, presence: true, uniqueness: { scope: :user_id }
+  validates :user_token, presence: true, uniqueness: true
```
brycemcd: This will allow each user to submit only one vote right? I think you can drop the uniqueness constraint here.

andrezafeu: Yes, but they can update their vote. If the same user tries to vote again, his/her old vote will be updated.

frontend/actions/submitVotes.js
```diff
@@ -45,6 +46,7 @@ export default function submitVotes({ votes }) {
         id,
         bill_summary_id,
         response: frontendToBackendVoteParams(vote.response),
+        user_token: vote.user_token,
```
e91096: you can pull this out above in line 44 (just like `id`) and then just have `user_token,` here!

frontend/components/VoteForm.jsx
```diff
@@ -1,13 +1,21 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { ensureObj } from '../helpers';
+import {
+  ensureString,
```
e91096: where do we use this?

frontend/components/CampaignShow.jsx
```diff
@@ -17,10 +17,18 @@ class CampaignShow extends Component {
   }
 
   componentWillMount() {
-    [
-      { resource: { singular: 'billSummary', plural: 'billSummaries' } },
-    ].forEach(args => this.props.fetchIfEmpty(args));
-    this.props.fetchCampaign(this.props.campaignId);
+    const queryString = require('query-string');
```
e91096: any reason we shouldn't require this at the top of the file instead of on mount?

frontend/actions/submitVotes.js
```diff
@@ -31,6 +31,7 @@ export function receiveVote(data) {
       vote: {
         ...data,
         response: backendToFrontendVoteParams(data.response),
+        user_token: data.userToken,
```
e91096: is the backend returning `userToken` camelcase?

frontend/components/VoteForm.jsx
```diff
@@ -1,13 +1,21 @@
...
+  ensureObj,
+} from '../helpers';
 
 const VoteForm = (props) => {
+
   const handleClick = (e) => {
     const vote = ensureObj(props.vote);
+    const queryString = require('query-string');
+    const parsed = queryString.parse(location.search);
+    const userToken = parsed.userToken;
```
e91096: is `VoteForm` ever called without being passed a vote which includes a user token?

frontend/components/CampaignShow.jsx
andrezafeu: Nope. I put it on top of the file now, thanks : )

frontend/actions/submitVotes.js
andrezafeu: oops

frontend/components/VoteForm.jsx
andrezafeu: Now it's on line 14 : )

brycemcd:
@andrezafeu have you rebased this on master since #40 was merged?

frontend/components/VoteForm.jsx
e91096: if `VoteForm` is passed a user token as part of its props, we don't need to parse the query string for the user token, right?

e91096:
@andrezafeu hey andreza! i pushed up some changes--let me know what you think. the setting of the user token to the current timestamp by default came out of a conversation with @brycemcd

frontend/components/VoteForm.jsx
andrezafeu: Thanks for the explanation : )

andrezafeu:
@e91096 your changes look good to me. I will just rebase it and make sure that everything is still working before I merge this!

PR #42, opened 2018-09-05
Adjusts header image to not repeat + cover image

brycemcd:
[Header Image Bug](https://trello.com/c/Ff3u6svA)

## Prerequisite PRs
None

## Dependent PRs
None

## Notes for Reviewer
Take a look in the browser in desktop and mobile mode. It _should_ render in the same way as Google Forms.


andrezafeu:
The page looks out of proportion in Safari (like if it had been zoomed out). I will take a look at that later.

brycemcd:
Could you provide a screenshot please? I don't have Safari.

andrezafeu:
First pic is Chrome and second is Safari.
I added <meta name="viewport" content="width=device-width, initial-scale=1.0"> to the head of application.html.erb but it didn't make any difference

Chrome
Safari

brycemcd:
Looks like the height was calculated differently in Safari. I can live with the differences :) What were you adding to the `<head>`?

andrezafeu:
Oh, it looks like my comment was incomplete.  I tried adding initial-scale=1 for width/height to the head of application.html.erb but it didn't make any difference. I think that's only for mobile

brycemcd:
AH, I see. That'll be a different problem a little later :)

PR #44, opened 2018-09-06
Add registration through invite

andrezafeu:
[Accept User Invitations](https://trello.com/c/bomO7FOx/13-accept-user-invitations)

## Prerequisite PRs

Generate Invite Tokens to Users (https://github.com/PocketLobby/constituents/pull/35) (merged)

## Dependent PRs
None

## Notes for Reviewer
Expected:
- If invite token is present, H1 header should update with the first name of the person that invited you
If registration is successful:
- Invite token param from `/register` path, if present, should be saved in db as `invited_by`. Otherwise, `invited_by` should be nil.
- New user should see his/her invite url in success page.


frontend/components/RegistrationForm.jsx
```diff
@@ -11,6 +11,9 @@ import {
   registerUser,
 } from '../actions/userActions';
 
+const match = window.location.href.match(/register\/([a-zA-Z\d-]+)/);
+const invitedBy = match ? match[1] : null;
```
e91096: i think it's a better pattern to save these kinds of things to component state

frontend/components/RegistrationForm.jsx
```diff
@@ -108,6 +115,10 @@ class RegistrationForm extends Component {
           errors.push('State should be two letters.');
         }
         break;
+      case 'invitedBy':
```
e91096: we don't need this case bc we catch `invitedBy` above in line 99!

frontend/components/RegistrationForm.jsx
```diff
@@ -121,7 +132,11 @@ class RegistrationForm extends Component {
     return (
       <div>
         <div>
-          <h1>Join</h1>
+          {invitedBy === null ? (
+            <h1>Join Pocket Lobby</h1>
+          ) : ( 
+            <h1>{invitedBy.charAt(0).toUpperCase() + invitedBy.substr(1).match('^[^-]+')} has invited you to join Pocket Lobby</h1>
+          )}
```
e91096: nice!

frontend/components/RegistrationForm.jsx
e91096: ah okay i understand. i'm not sure if it makes sense here as this is more for user-input validation errors. question: does a user need to have an invite token to register?

andrezafeu: I agree that we should allow them to register. I rebased and updated the code to validate the invited_by only in the back-end

app/models/user.rb
```diff
@@ -33,4 +36,10 @@ def generate_invite_token
     "#{self.first_name}-#{self.last_name}-#{random_alphanum}".downcase
   end
 
+  def validate_invited_by
+    return if self.invited_by == nil
+    # REVIEW regex to make sure entire string matches
```
e91096: is this still todo?

frontend/components/RegistrationForm.jsx
```diff
@@ -11,6 +11,7 @@ import {
   registerUser,
 } from '../actions/userActions';
 
+
```
e91096: extra space

frontend/components/RegistrationForm.jsx
```diff
@@ -86,10 +99,11 @@ class RegistrationForm extends Component {
       state: 'State',
       USCitizen: 'US Citizen',
       zip: 'Zip',
+      invitedBy: 'Invited By',
```
e91096: not needed

frontend/routes.jsx
```diff
@@ -22,7 +22,7 @@ const routes = [
   { path: /^\/admin\/campaigns\/(\d+)$/, action: () => <CampaignForm /> },
   { path: /^\/login/, action: () => <LoginForm /> },
   { path: /^\/register$/, action: () => <RegistrationForm /> },
-  { path: /^\/register\/([a-zA-Z\d-]+)$/, action: () => <RegistrationForm /> },
+  { path: /^\/register\/(.+)$/, action: () => <RegistrationForm /> },
```
e91096: why not make it `(.*)` and just have one path that routes to the `RegistrationForm`?

frontend/components/SuccessStub.jsx
```diff
@@ -7,6 +7,9 @@ import { logoutUser } from '../actions/userActions';
 const SuccessStub = (props) => {
   return (
     <div>
+      <h1>
+        Thanks for registering!
+      </h1>
       <h2>
         Your invite URL is {window.location.origin}/register/{props.currentUser.invite_token}
```
e91096: i'm not seeing an invite-token after i register (as a new user)

frontend/routes.jsx
andrezafeu: I tried but this didn't work for me. I'm getting an error in the async function in the router, but can't figure out what's wrong with it. Do you mind taking a look at that when you review this?

app/models/user.rb
andrezafeu: oops, that's done

frontend/components/RegistrationForm.jsx
andrezafeu: removed : )

frontend/components/SuccessStub.jsx
andrezafeu: fixed

e91096: works!

frontend/components/RegistrationForm.jsx
```diff
@@ -30,6 +30,17 @@ class RegistrationForm extends Component {
     this.handleSubmit = this.handleSubmit.bind(this);
   }
 
+  componentDidMount() {
+    const href = window.location.href;
+    if (href.includes('inviteToken')) {
```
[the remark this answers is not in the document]
andrezafeu: It will be true if the inviteToken is present in the url. Otherwise, it will be false. But I just noticed that I don't need this `if` statement. That's what you had in mind when you made this question? I refactored the code : )

andrezafeu:
@e91096 regarding the invite token: before I was also getting the invite token from the url, but I was using `.match`. For some reason, that piece of code stopped working when I rebased the branch. So I updated it to this one.
For the success page I think that, since the invite token used is generated in the BE, it makes more sense to serve from the BE than from the url. The url would look 'cleaner' and it doesn't let the user manipulate the value. Let me know if you agree!

frontend/components/RegistrationForm.jsx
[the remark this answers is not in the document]
e91096: hmmm i mean is it ever the case that the url contains the string `'inviteToken'`? the one that we present to the user looks like this: `http://localhost:3000/register/test-user-iswjp3mian` and not `http://localhost:3000/register/inviteToken=test-user-iswjp3mian`

e91096:
yup that's a great idea!
i was more concerned that the url we provide to the users doesn't match the format of the url we're expecting to parse when we're in the registration form?

andrezafeu:
@e91096 Ohhhhhh I see now. I misunderstood your comment before!

andrezafeu:
@e91096 when I defined the route in the FE, I thought it would be better to accept any content after the `register/` part of the path. So if the url gets broken in any way, this won't prevent the user from being registered. It's more a less what you suggested [here](https://github.com/PocketLobby/constituents/pull/44#discussion_r217927889) . If there's any change you think it's worth doing, could you be more specific? Or were you just double checking that the route works? Let me know!

PR #47, opened 2018-09-10
BUG: submit votes

brycemcd:
This was a production bug. Removes dependency on users being
logged in to submit votes. NOTE: after deploy, create user
with an id of -99. It's a magic number.

[Trello Ticket](https://trello.com/c/yr9DDkii)

## Prerequisite PRs
None

## Dependent PRs
None

## New ENV Vars
AUTH_ENABLED - not required. If not present, then there is no harm to the system (and it's been tested). 

## Notes for Reviewer

Please note that the previous code must not have been tested very well prior to being released. This code wraps authentication in an environment variable so that we can _not_ depend on auth in the "No Mo' Google Forms" release.


brycemcd:
I created the fake user in prod. Merging and :shipit:

PR #48, opened 2018-09-10
Bill Summary/Opinion UI tweaks

brycemcd:
I noticed some tweaks that could be made in the UI to make it more consistent with the presentation our users have seen for the last few months. This PR is an OPPREF (outside usual sprint work) to get some small data, markup and styling tweaks into prod.

## Prerequisite PRs
None

## Dependent PRs
None

## New ENV Vars
None

## Notes for Reviewer
Take a look at the campaign landing page. All changes _should_ be isolated to that page. Couldn't hurt to look at some other pages too just to be sure.


frontend/components/VoteForm.jsx
```diff
@@ -23,16 +23,26 @@ const VoteForm = (props) => {
...
-      </div>
-      <form className = "form-description-element">
+      </span>
+
+      <span className = "form-description-element full-bill-url">
+        (full bill:
+          <a
+            className = "bill-summary-url"
+            href={props.billSummary.full_bill_url}
+            target="_blank"> {props.billSummary.full_bill_url} </a>)
```
andrezafeu: This is working as a relative path right now (bill url is added after our site url). It should be an absolute path. I added a new issue in Trello's backlog.

frontend/stylesheets/forms.css
```diff
@@ -47,10 +47,11 @@ p {
 
 .form-info {
   width: 80%;
-  max-width: 520px;
+  max-width: 746px;
```
andrezafeu: it looks much better with this width : )

frontend/stylesheets/forms.css
```diff
@@ -76,22 +77,53 @@ p {
...
 }
 
 .form-navigation {
-  margin-top: 30px;
-  padding-left: 25px;
-  padding-right: 25px;
+  margin: 30px 0;
 }
 
 .form-button {
```
andrezafeu: The way this is styled, if we add navigation buttons or other buttons (like cancel, or back), they will stack on top of each other, instead of side by side. But we can deal with that later if that's the case.

frontend/components/VoteForm.jsx
brycemcd: Yeah, that'll happen if you don't include http(s) when developing the campaign. I'll see about updating the seed.rb file

frontend/stylesheets/forms.css
brycemcd: agreed. I think this should be a flex box personally but I'm trying to solve the problem in front of me more that trying to create a perfect solution. IMHO, it's intuitive and easy to change this way. I'd prefer to leave it the way it is unless you have a strong opinion.

frontend/components/VoteForm.jsx
andrezafeu: Cool, I didn't know that

andrezafeu:
LGTM!!!

PR #50, opened 2018-09-11
Refactor Campaign API

brycemcd:
Refactoring the campaign API response to include our current needs and support refactoring the FE a bit

[Trello Ticket](TRELLO_TICKET_URL)

## Prerequisite PRs
None

## Dependent PRs
None

## New ENV Vars
None

## Notes for Reviewer

Run a few curl commands to make sure you get back what you expect


app/controllers/api/campaigns_controller.rb
```diff
@@ -24,11 +24,14 @@ def create
...
+
     if campaign
-      render json: campaign.as_json
-      # render json: CampaignPresenter.new(campaign)
+      render_json('campaign', campaign.as_json(user_token: params[:userToken]),
+                                       status_code: 200)
     else
-      render body: nil, status: 404
+      render_json('campaign',
+                  ['That campaign could not be found'],
```
andrezafeu: I'm not getting this message back. It fails on line 26 and this code doesn't seem to run

docs/api/campaigns.md
```diff
@@ -0,0 +1,72 @@
...
+curl -v http://localhost:3000/api/campaigns/1?userToken=foo
+```
+
+### Success Response
+
+Status code: 200
+
+```json
+{ "action": "campaign",
+  "success": false,
```
andrezafeu: shouldn't this be true?

docs/api/campaigns.md
```diff
@@ -0,0 +1,72 @@
...
+  "obj" : {
+    "id" : 4,
+    "name" : "good campaign",
+    "bill_summaries" : [
+      { "id" : 1,
+        "full_bill_url" : "https://www.fakecongress.org/123",
```
andrezafeu: just a nitpick: the full_bill_url actually gets returned after the bill_number

spec/controllers/campaigns_spec.rb
```diff
@@ -3,4 +3,38 @@
...
+        expect(response.body).to have_expected_success_json_structure
+      end
+    end
+
+    describe "when id DOES NOT exist" do
+      it 'returns 400 and error json' do
+        expect(Campaign).to receive(:find).and_return(nil)
+        get :show, params: {id: "bunk", userToken: 'foo'}
+        expect(response.status).to eql(404)
+        expect(response.body).to have_expected_failed_json_structure
```
andrezafeu: I think this test should also check if the message 'That campaign could not be found' gets returned

docs/api/campaigns.md
brycemcd: well, only if you want it to be accurate. Great catch!

docs/api/campaigns.md
brycemcd: great nitpick but in ruby hashes, order is not guaranteed. Our consumers shouldn't expect ordering to be consistent. Mayhaps I can make this more clear in the docs.

spec/controllers/campaigns_spec.rb
brycemcd: I think I agree with you to an extent. Do you feel like the actual message is important? I'm checking that there _is_ a message but not what it is. The reasoning being that if we change the message, the test will fail; it's kind of brittle. I'd like to hear your thoughts on it though.

andrezafeu: I think the way you did now looks good.  I thought about that because I was getting an error that was not being caught by the tests.

docs/api/campaigns.md
andrezafeu: Fun fact: I got curious and googled it. It's ordered according to insertion order since ruby 1.9  : )
But I think the explanation you added to the docs is sufficient.

brycemcd: nice! Thanks for getting the right answer to this. It proves that looking it up is better than using my outdated recollection. Nice work.

PR #49, opened 2018-09-10
bug add missing migration file, refactor votes into conventional response

andrezafeu:
[Migration that adds Vote#user_token](https://pocketlobby.slack.com/archives/C7Q12JY21/p1536604881000100)

## Prerequisite PRs
None

## Dependent PRs
None

## New ENV Vars
None

PR #52, opened 2018-09-11
Changing votes API to return our conventional response

brycemcd:
First step in upserting votes via the API

## Prerequisite PRs
#49

## Dependent PRs
None

## New ENV Vars
None

## Notes for Reviewer
Note the docs and run the curl commands to make sure they're correct. Pay extra special attention to the tests I've removed. I expect that it should be fine but I may have just gone on a frenzy deleting code.


andrezafeu:
This looks great! The only thing is that the success status code I get is 201 (not 200 like the docs). Also, maybe there should be some sort of validation on the user_token value? I was able to use characters like @ and #

brycemcd:
201 should be the right response. I'll update the docs.

I think Postel's law (kind of) applies here: We can be liberal in  what we accept but conservative in what we return. Personally, I don't think we need to do validation on the user token. It either matches or it doesn't. 

If we validate the token then we have to change it when we change the token paradigm later.

PR #51, opened 2018-09-11
fix user-specific vote submission; fix redirect after vote submission

e91096:
related - [Trello Ticket](https://trello.com/c/NnczLbcO/53-bug-if-forms-are-submitted-but-not-saved-success-path-is-followed)

## Prerequisite PRs
- [x] https://github.com/PocketLobby/constituents/pull/50
- [x] https://github.com/PocketLobby/constituents/pull/49

**NOTE** this branch is based on both the above branches. Please select the last 3 commits when reviewing (https://github.com/PocketLobby/constituents/pull/51/files/632f7ecd09bde45305f7df543529ca8c105bd76f..ef4cb2de03505500955ae0da39cdecd53cf876ab)

## Dependent PRs
n/a

## New ENV Vars
n/a

## Notes for Reviewer
Things that have been touched:
- campaign fetching, and by extension vote fetching
- vote fetch/submission

Things that should still work:
- a campaign show page should populate correctly with the campaign and its bill summaries
- votes for those bill summaries should be answered appropriately
- a user's votes should be saved

Things that should _now_ work:
- a different user should be able to vote without overwriting any other user's votes

PR #54, opened 2018-09-13
bug process env

brycemcd:
Redirecting after vote submission wasn't working and it would not have worked in prod because we need to read in an env var from the host. This PR allows fixes that.

**Note** This PR is meant to be sequenced after #51 . Merge that one in, then this one

## Prerequisite PRs
#51

## Dependent PRs
None

## New ENV Vars
None new, but uses `VOTE_REDIRECT` that was previously depended

## Notes for Reviewer
Test it by voting. Set `VOTE_REDIRECT` https://www.example.com/it-works and note that you're directed there when your vote is complete


brycemcd:
it was easier to rebase and push to #56

PR #56, opened 2018-09-14
BUG: process.env didn't work

brycemcd:
Fixes bug in making env vars work in the FE. This is a rebased version of #54 

## Prerequisite PRs
None

## Dependent PRs
None

## New ENV Vars
None new, but uses `VOTE_REDIRECT` that was previously depended

## Notes for Reviewer
Test it by voting. Set `VOTE_REDIRECT` https://www.example.com/it-works and note that you're directed there when your vote is complete


andrezafeu:
It works like a charm ; )
Also, great addition to the README

PR #57, opened 2018-09-14
BUG: fix compilation error + add some VERY basic tests in VoteQueuer

brycemcd:
The app is crashing in prod because of a careless error _I_ made.

## Prerequisite PRs
None

## Dependent PRs
None

## New ENV Vars
None

## Notes for Reviewer
Run the tests. (or check CI.) As long as they run, and pass, this error should be eliminated. The file wasn't being loaded in the dev and test environments previously. I'm not sure why this is but there was a difference between those envs and production. I added very basic tests so that the file is loaded in the dev/test environments.


app/models/vote_queuer.rb
```diff
@@ -18,15 +17,7 @@ def write_to_queue
...
-
-    sqs.send_message(
-      queue_url: ENV['AWS_QUEUE_URL'],
-      message_body: {
-        user_token: @user_token
-        bill_opinions: bill_opinions,
-        timestamp: timestamp,
-      }.to_json
-    )
+    send_message(bill_opinions, bill_opinions)
```
andrezafeu: I think the second argument should be the timestamp!

brycemcd: :face_with_head_bandage: wow that's sloppy of me. NICE catch

PR #59, opened 2018-09-16
WIP - After registration, add user to MailChimp

andrezafeu:
[After registration, add user to MC](https://trello.com/c/pq3DRXNU/42-after-registration-add-user-to-mc)

## Prerequisite PRs
None

## Dependent PRs
None

## New ENV Vars
`MAILCHIMP_API_KEY`
`MAILCHIMP_LIST_ID`

## Notes for Reviewer
I'm not sure why the CI is failing


app/models/user.rb
```diff
@@ -25,11 +30,31 @@ def api_return
     }
   end
 
+  def md5_hashed_email
+    Digest::MD5.hexdigest(self.email)
+  end
```
brycemcd: Is this method in the right place? It really tingles my spider sense to see `md5` in a `User` model :)

If it's only used for Mailchimp reasons, then I think a more appropriate place for it to live would be the Mailchimp class.

app/models/user.rb
```diff
@@ -25,11 +30,31 @@ def api_return
...
+
+  def upsert_user_to_mailchimp
+    Mailchimp.new().upsert_user(self)
+  end
+
+  def citizen_yes_or_no
+    self.us_citizen ? 'Yes' : 'No'
+  end
```
brycemcd: This is a mailing list concern, not a user model concern. I'm happy to be proven otherwise, but I think it would be best in the Mailchimp class.

A couple of guiding principles to think about as you work on this is: 
* if we changed mail providers, would this still make sense?
* Even though it interacts with a user, is it relevant to manipulating a user object or storing it to the DB?

IMHO, I think the answer to both those questions is "no." If we moved from Mailchimp to Monkey Mailer (or something) then there's a good chance we wouldn't use "yes" and "no" for the us_citizen field. We'd want that data mapping in a Monkey Mailer class. It also isn't stored in the database. The value is mapped from the boolean value we care about to the value that we can store in our mail provider.

There's certainly nothing _wrong_ with this code. I simply think it could be factored a bit better.

app/services/mailchimp.rb
```diff
@@ -0,0 +1,40 @@
...
+          DIST_NUM: user.congressional_district.to_s,
+          STATE: user.state.to_s,
+          USER_TOKEN: user.user_token.to_s,
+          US_CITIZEN: user.citizen_yes_or_no,
+          ZIP_CODE: user.zip.to_s
+        }
+      }
+    )
+  rescue Gibbon::MailChimpError => e
+    raise e
```
brycemcd: Why are you rescuing this error only to re-raise it?

db/migrate/20180916004538_add_user_token_to_users.rb
```diff
@@ -0,0 +1,6 @@
+class AddUserTokenToUsers < ActiveRecord::Migration[5.1]
+  def change
+    add_column :users, :user_token, :string, null: false
+    add_index :users, :user_token, unique: true
```
brycemcd: nice

spec/services/mailchimp_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  describe "Mailchimp" do
+    let(:user) { FactoryBot.build(:user) }
+
+    it "initializes mailchimp and upserts user to API" do
+      authorization_header = "apikey:#{ENV['MAILCHIMP_API_KEY']}-#{ENV['MAILCHIMP_DATACENTER']}"
+
+      stub_request(:put,
+                   "https://#{ENV['MAILCHIMP_DATACENTER']}.api.mailchimp.com/3.0/lists/#{ENV['MAILCHIMP_LIST_ID']}/members/#{user.md5_hashed_email}")
+      .with(
+        body: "{\"email_address\":\"#{user.email}\",\"status\":\"subscribed\",\"merge_fields\":{\"FNAME\":\"#{user.first_name}\",\"LNAME\":\"#{user.last_name}\",\"DIST_NUM\":\"#{user.congressional_district}\",\"STATE\":\"#{user.state}\",\"USER_TOKEN\":\"#{user.user_token}\",\"US_CITIZEN\":\"#{user.citizen_yes_or_no}\",\"ZIP_CODE\":\"#{user.zip}\"}}",
```
brycemcd: Would this be easier to write as a hash and then cast it to JSON?

i.e.
```ruby
{ body: {"email_address": user.email}.to_json }
```

spec/spec_helper.rb
```diff
@@ -1,3 +1,5 @@
+require 'webmock/rspec'
+WebMock.disable_net_connect!(allow_localhost: true)
```
brycemcd: good job figuring this out on your own!

app/models/user.rb
```diff
@@ -25,11 +30,31 @@ def api_return
     }
   end
 
+  def md5_hashed_email
+    Digest::MD5.hexdigest(self.email)
+  end
+
+  def upsert_user_to_mailchimp
+    Mailchimp.new().upsert_user(self)
```
brycemcd: This is a nitpick but you could refactor this to make it a class method and pass in the `user_id`. Then, lookup the user in the Mailchimp class and pass the instance there.

It's a fairly small change but would help decoupling IMHO. This would be reflected in the tests as well. You could easily stub the class method in tests where the `after_create` method is called.

app/services/mailchimp.rb
```diff
@@ -0,0 +1,40 @@
+class Mailchimp
+  def initialize
+    if ENV['MAILCHIMP_API_KEY'] && ENV['MAILCHIMP_DATACENTER']
+      # TODO implement logger option
+      # Ruby Logger.new is used by default, but it can be overrided by passing: logger: MyLogger.new)
+      @mailchimp = Gibbon::Request.new(api_key:         ENV['MAILCHIMP_API_KEY'] +
+                                                        '-' +
+                                                        ENV['MAILCHIMP_DATACENTER'],
```
brycemcd: The Mailchimp API key in the admin is structured like: `abc123-us16`. It seems that the datacenter is already part of the API key. Is there a good reason to factor it out like this? Would we ever call a different data center with the same key?

app/services/mailchimp.rb
andrezafeu: Your question makes sense lol. I will log it

app/models/user.rb
andrezafeu: I agree with you. Thanks for making these suggestions : )

andrezafeu:
thanks for all the suggestions @brycemcd ! I agree with all of them, it does make more sense to have some of those methods inside the mailchimp class. I added a couple more tests that I had ready.

app/models/user.rb
andrezafeu: done

app/services/mailchimp.rb
andrezafeu: I guess not. I refactored the code!

brycemcd:
Glad to hear it! It's more important to me that you understand the _why_ of the suggestions rather than blindly copying/pasting code. Glad we're on the same page!

This PR LGTM. :ship: :it: real good

PR #60, opened 2018-09-17
Fix incorrect redirect when a bill summary fails to save

chrisruenes1:
1. Esther already fixed this error for votes in another branch, before the card existed.
2. Move the rerdiect out of the component, and put instead into the success functions
for the actions
3. Returrn 422 status codes in bill_summaries when there is an errorr

[Trello Ticket](https://trello.com/c/NnczLbcO/53-bug-if-forms-are-submitted-but-not-saved-success-path-is-followed)


app/controllers/api/bill_summaries_controller.rb
```diff
@@ -34,7 +34,7 @@ def update
       if bill_summary.update(permitted_params)
         render json: bill_summary
       else
-        render json: bill_summary.errors.full_messages
+        render json: bill_summary.errors.full_messages, status: 422
```
brycemcd: would it be a big deal to refactor these two methods to use our consistent `render_json` (in application controller) helper method?

chrisruenes1: @brycemcd pushed these updates!

chrisruenes1:
@brycemcd let me know if this second batch of changes (updating to the standard json response) looks ok to you and I'll merge it in!

brycemcd:
LGTM :shipit:

PR #61, opened 2018-09-17
Render consistent errors

chrisruenes1:
[Trello Ticket](https://trello.com/c/RS5zy1lJ/40-design-10)

In the spirit of standardizing the look of the frontend, I refactored the error modal into a connected component that is a direct child of the Layout component. It now is always on the page, and slices of state determine whether it shows or hides. The messages are also moved into the store. For now, changes to the error state are only made: 

1) when an error status is returned from the server
2) when the user exits out of the modal

Still need to style better, esp in terms of layout, and add this logic outside of bill summary methods.

## Prerequisite PRs
https://github.com/PocketLobby/constituents/pull/60


frontend/actions/userActions.js
```diff
@@ -31,14 +33,18 @@ export function loginUser(params, successFn, failFn) {
...
+        } else {
+          dispatch(showErrors(['Your email and password combination was not correct.']));
+        }
       },
     });
   };
 }
 
-export function logoutUser(successFn, failFn) {
+export function logoutUser(successFn = _.identity, failFn) {
```
e91096: oooh what's the `_.identity` for?

frontend/actions/userActions.js
```diff
@@ -50,14 +56,18 @@ export function logoutUser(successFn, failFn) {
         dispatch(logoutCurrentUser());
         successFn();
       },
-      error: (data) => {
-        failFn();
+      error: () => {
+        if (_.isFunction(failFn)) {
+          failFn();
+        } else {
+          dispatch(showErrors(['Sorry, something went wrong. Please try to log out again.']));
```
e91096: 😍

frontend/components/ErrorModal.jsx
```diff
@@ -1,14 +1,46 @@
...
-    </section>
-  );
+  const errors = props.errors.map(error => (<div className="form-validation-error" key={hash(error)}> {error} </div>));
+  return props.showModal ? (
+    <div>
+      <button
+        className="dismiss"
+        onClick={props.dismissErrors}
+      >
+        x
```
e91096: `&times;`?

frontend/components/ErrorModal.jsx
```diff
@@ -1,14 +1,46 @@
...
-  });
-  return (
-    <section className="form-validation-error">
-      {errors}
-    </section>
-  );
+  const errors = props.errors.map(error => (<div className="form-validation-error" key={hash(error)}> {error} </div>));
+  return props.showModal ? (
+    <div>
+      <button
```
e91096: great idea

frontend/components/Layout.jsx
```diff
@@ -16,12 +16,15 @@ export default function Layout(props) {
...
 
   return (
-    <div className="group">
-      {showNavbar(window.location.href) &&
-        <Navbar />
-      }
-      <div>
-        {props.children}
+    <div>
+      <ErrorModal />
```
e91096: should the error modal go under the navbar?

frontend/actions/userActions.js
chrisruenes1: I'm not sure how I feel about this now! I added it just as default value so that we don't always need to supply successFn. Or in case the question was about `_.identity`, that just returns exactly what is passed, it's like a noOp!

frontend/stylesheets/errors.css
```diff
@@ -9,3 +9,18 @@ div.form-validation-error {
...
+  margin: 6px 12px;
+  background: transparent;
+  font-size: 20px;
+  border: none;
+  color: maroon;
+  cursor: pointer;
+}
+
+.dismiss:focus {
+  outline: none;
```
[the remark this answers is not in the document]
chrisruenes1: Yeah I did not like that blue box!

frontend/components/ErrorModal.jsx
chrisruenes1: Do you remember how to render html elements in react? I tried this and it rendered literally, and then tried with quotes and the linter complained, and when I followed the link, didn't really make the complaint make more sense.

frontend/components/Layout.jsx
chrisruenes1: @e91096 when I did this, putting it under the navbar, I think, had it rendering weird, but I'll bet that's not the case with the new styling! I'll double check and move it below if so

chrisruenes1:
@e91096 yay thanks for the review! Hehe the reason it doesn't clear on navigate is cause uhhh I forgot I was going to do that. I'll take a look at how tomorrow! And will look into the bug on login page

frontend/stylesheets/nav_bar.css
```diff
@@ -3,10 +3,6 @@ html, body, main, div.layout {
   font-family: 'Open Sans', sans-serif;
 }
 
-nav.nav-container + div {
```
e91096: oh okay so this is so the div under the navbar takes up the full height of the screen. otherwise you get something like this: 
screen shot 2018-10-02 at 11 05 20 pm
which is no good... 
maybe you could use a `:not` here to exclude error modals?

frontend/actions/userActions.js
e91096: ohhh interesting. so we don't have to check if successFn exists before invoking... seems reasonable to me!

frontend/components/Layout.jsx
e91096: also, not that this is not rendered within the form's margins, it should probably have its own margin! currently, it kiiiinda looks like an extension of the navbar
screen shot 2018-10-02 at 11 12 29 pm

frontend/components/ErrorModal.jsx
e91096: hmmm it works for me when i just stick it right in there in place of the `x`
```
<button
  className="dismiss"
  onClick={props.dismissErrors}
>
  &times;
</button>
```

frontend/index.jsx
```diff
@@ -24,11 +26,9 @@ function renderComponent(component) {
 }
 
 function render(location) {
+  store.dispatch(dismissErrors());
```
e91096: nice

frontend/components/Layout.jsx
chrisruenes1: oooh good call!

chrisruenes1:
@e91096 Remaining changes made! Let me know if there's anything else, otherwise I'm gonna Carnival Cruise it!

PR #63, opened 2018-09-19
Password Reset BE

brycemcd:
[Trello Ticket](https://trello.com/c/PwgLgsLr)

+ [x] update docs to describe endpoints
+ [x] configure routes to use Devise
+ [x] implement endpoints

## Notes for Reviewer
To test it:
- Go to http://localhost:3000/api/password/new
- Add an email that's not in the database. Verify that you get a 404.
- Go back and add an email that is in the database. You should get a 201 and see the reset password link in the logs
- Test curl request


config/routes.rb
```diff
@@ -10,7 +10,8 @@
                },
                controllers: {
                  sessions: 'api/sessions',
-                 registrations: 'api/registrations'
+                 registrations: 'api/registrations',
+                 passwords: 'devise/passwords',
```
andrezafeu: Hi there, any special reason for this to follow the pattern `devise/passwords` instead of `api/passwords`?

app/controllers/api/passwords_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+  def create
+    self.resource = resource_class.send_reset_password_instructions(resource_params)
+    yield resource if block_given?
+
+    if successfully_sent?(resource)
+      render_json('password_reset', { email: resource.api_return[:email] },
+        status_code: 201)
+    else
+      render_json('password_reset', resource.errors.full_messages,
+      status_code: 422)
```
brycemcd: shouldn't this be a 404?

config/routes.rb
```diff
@@ -10,7 +10,8 @@
                },
                controllers: {
                  sessions: 'api/sessions',
-                 registrations: 'api/registrations'
+                 registrations: 'api/registrations',
+                 passwords: 'api/passwords',
```
brycemcd: I had made this `devise/passwords` so that we could use the built in devise code without copying/pasting the controller code into our own controller. That was the impetus for creating a jbuilder template since we wouldn't be able to override the return value of Devise's controller but _could_ return our own template.

docs/api/password_reset.md
```diff
@@ -0,0 +1,38 @@
...
+```bash
+# succeeds if user has been created previously
+curl -d api_user[email]=valid@email.com http://localhost:3000/api/devise/password.json
+```
+
+### Success Response
+
+Status code: 201
+
+{
```
brycemcd: needs to be wrapped in ```json

andrezafeu: thanks. done

PR #64, opened 2018-09-20
mailchimp tests

brycemcd:
Fix tests on the MC branch

## Notes for Reviewer
Note the preference for `build_stubbed` over `create` in the factories. It saves a trip to the database. (likely, it saves many trips to the database setting/releasing transaction locks.

I also mocked the call to `Mailchimp` in the user model. It was helpful that was moved to a class method :) Ideally, we would use dependency injection but I didn't see a trivial way to do that while also maintaining robust tests and avoiding an "it worked in dev" kind of production bug.


andrezafeu:
This looks great, thanks!

PR #70, opened 2018-09-24
WIP - API definition

brycemcd:
Build an API endpoint to know the current live campaign

## Prerequisite PRs
None

## Dependent PRs
Not a dependency per-se, but a good-to-have: #33 
For now, I'm just getting the last updated campaign and pretending that will work as the active one. For the purposes of our production app, this will be fine until #33 is merged. We can refactor (very easily!) after that.

## New ENV Vars
None


config/routes.rb
```diff
@@ -13,13 +13,15 @@
...
     resources :bill_summaries, only: [:index, :show, :create, :update, :destroy]
-    resources :campaigns, only: [:index, :show, :create, :update, :destroy]
     resources :votes, only: [:create, :update]
   end
 
   # NOTE: we don't want bots trolling the domain and figuring out they can
   # CRUD bill summaries and campaigns
   root to: redirect("https://www.pocketlobby.org")
-  get '*path', to:"static_pages#root"
+  #get '*path', to:"static_pages#root"
```
e91096: intentional?

brycemcd: oops!

PR #71, opened 2018-09-25
BUG: fixes null constaint exception in production

brycemcd:
When adding a null constraint, one must ensure existing
records are populated

## Prerequisite PRs
None

## Dependent PRs
None

## New ENV Vars
None

## Notes for Reviewer
This is an urgent production issue


db/migrate/20180916004538_add_user_token_to_users.rb
```diff
@@ -1,6 +1,30 @@
...
+    remove_column :users, :user_token
+  end
+
+  private
+
+  def init_all_user_tokens
+    User.reset_column_information
+
+    User.all.each do |u|
+      u.send(:set_user_token)
```
e93285: silly question: what does this set the user token to?

e91096: silly question: what does this set the user token to?

brycemcd: There are no silly questions, only silly people.

`u.send(:set_user_token)` calls a private method on the user object, `set_user_token`, which "does the right thing" by setting a random string in our user token style if one has not been set or immediately returning if a user_token is already set.

Normally, this is a code smell (nice work!) because we don't want to be calling private methods on user objects. IMHO, it's okay _in this instance_ because it's a run-once migration setting a new property.

e91096: aha! 👍

brycemcd:
woot! Thanks!

PR #73, opened 2018-09-26
Create Password Reset Form

chrisruenes1:
[Trello Ticket](https://trello.com/c/C5gLPH8I/57-a-constituent-should-be-able-to-reset-their-password-be-fe)

## Notes for Reviewer
I renamed `ResetPasswordForm to RequestPasswordResetForm` to differentiate it from the form you get redirected to after the email. I'm open to better names! Also, I didn't do it with git move, so the diffs on the `ResetPasswordForm` are not really diffs per se...They are code from a new file replacing line by line code from an old file. If this gets too confusing, let me know and I can do `git mv` instead on another branch.

I also added error modal behavior to both forms.

I removed the functionality for checking if no errors exist before submit from the `RequestPasswordResetForm` because it was failing silently in the new implementation, once I took the error modal out of that component. I'm also open to alternative suggestions here. I think it would at least be good to recreate the submit disable functionality.


frontend/components/PasswordResetForm.jsx
```diff
@@ -1,93 +1,105 @@
...
+    this.props.resetPassword({
+      userPasswordResetToken: this.props.userPasswordResetToken,
+      password: this.state.password,
+      passwordConfirm: this.state.passwordConfirm,
+    });
+  }
 
+  render() {
     return (
-      <div className="form-container">
```
e91096: i think there might have been changes to this file made since you renamed it--maybe we should re-copy it over from master?

e91096: (but obv update all of the passwordreset -> requestpasswordreset again)

frontend/components/PasswordResetForm.jsx
```diff
@@ -1,93 +1,105 @@
...
+
 import { connect } from 'react-redux';
 import * as _ from 'underscore';
-import ErrorModal from './ErrorModal';
+import PropTypes from 'prop-types';
+
+
 import Input from './shared/Input';
+import '../stylesheets/reset.css';
+import '../stylesheets/forms.css';
```
e91096: i've been using `login_form.css` instead of `forms.css` for all the non-vote forms! i think it'd make sense to use that here also?

frontend/components/PasswordResetForm.jsx
```diff
@@ -1,93 +1,105 @@
...
 
 const mapDispatchToProps = dispatch => ({
-  resetPassword: (params, redirect, handleFail) =>
-    dispatch(resetPassword(params, redirect, handleFail)),
+  resetPassword: params => dispatch(resetPassword(params)),
 });
 
-const mapStateToProps = state => ({});
+const mapStateToProps = () => {
+  const userTokenRegex = /\?userPasswordResetToken=(.+)/;
```
e91096: silly question: how do we know the link they're sent looks like this?

chrisruenes1: @brycemcd can you confirm that the url looks like this?

chrisruenes1:
@e91096 thanks for the review! I think that since your approved PR has changes to the `PasswordResetForm` (which I renamed), it may make sense to merge yours in first, then I can rebase and get this to where it should be commit.

chrisruenes1:
also, @brycemcd Ester and I were trying to parse how dependent our PRs are each other? It seems like the password reset flow is essential for enabling auth, but we weren't sure whether we may currently have a way for users to send invite links without being properly logged in?

brycemcd:
@chrisruenes1 / @e91096 We need to have password reset in order to enable auth. It's a precondition to allowing invites to begin happening.

frontend/components/PasswordResetForm.jsx
chrisruenes1: I think it's a good inclination to consider this an `edit` link following the rails convention, but how would we feel about removing the `api` part of the link, since it's not really hitting that?

frontend/actions/userActions.js
```diff
@@ -85,8 +86,8 @@ export function registerUser(params, successFn = _.identity, failFn) {
   };
 }
 
-export function resetPassword({ email }, successFn, failFn) {
-  return function resetPasswordThunk(dispatch) {
+export function requestResetPassword({ email }, successFn, failFn) {
```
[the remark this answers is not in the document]
chrisruenes1: oooh good catch!

frontend/components/PasswordResetForm.jsx
chrisruenes1: 👍

frontend/components/PasswordResetForm.jsx
```diff
@@ -1,93 +1,105 @@
...
 class PasswordResetForm extends Component {
   constructor(props) {
     super(props);
-
     this.state = {
-      email: '',
-
-      displayConfirmation: false,
-      errors: [],
-      submitDisabled: false,
```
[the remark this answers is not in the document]
chrisruenes1: So, I took this stuff out when I was converting that form to using the error modal, but I may have gotten a little slash-happy! In any case, I've copied the file over directly from master now, so I'll make sure I don't delete any functionality this time around!

frontend/components/PasswordResetForm.jsx
chrisruenes1: 👍

frontend/components/RequestPasswordResetForm.jsx
```diff
@@ -0,0 +1,68 @@
+import React, { Component } from 'react';
```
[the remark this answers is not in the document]
chrisruenes1: Yup! I did a git move this time and I think the diffs should be less wacky

frontend/components/PasswordResetForm.jsx
```diff
@@ -1,93 +1,105 @@
 import React, { Component } from 'react';
+
 import { connect } from 'react-redux';
 import * as _ from 'underscore';
-import ErrorModal from './ErrorModal';
+import PropTypes from 'prop-types';
+
+
```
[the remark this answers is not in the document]
chrisruenes1: ah thanks!

frontend/components/PasswordResetForm.jsx
chrisruenes1: Gonna leave it in for now so we can get this merged and unblock the release, but we can have this convo later if we want!

chrisruenes1:
@e91096 this should be ready for re-review! Annoyingly, I tried using `git mv` to prevent these "diffs" that are actually a new file being created, but it looks like it's still presenting them that way

frontend/routes.jsx
```diff
@@ -21,8 +22,9 @@ const routes = [
   { path: /^\/campaigns\/(\d+)$/, action: () => <CampaignShow /> },
   { path: /^\/admin\/campaigns\/(\d+)$/, action: () => <CampaignForm /> },
   { path: /^\/login$/, action: () => <LoginForm /> },
-  { path: /^\/password-reset$/, action: () => <PasswordResetForm /> },
+  { path: /^\/api\/password\/edit\/(.+)$/, action: () => <PasswordResetForm /> },
```
e91096: if somehow they navigate themselves to `/api/password/edit` without the code, they'll get a rails error page instead of the main content error page.. that's probably fine for now?

frontend/components/PasswordResetForm.jsx
```diff
@@ -1,103 +1,104 @@
...
-import PropTypes from 'prop-types';
 import * as _ from 'underscore';
+import PropTypes from 'prop-types';
+
 
 import Input from './shared/Input';
-import { resetPassword } from '../actions/userActions';
-import { showErrors } from '../actions/errorActions';
+import '../stylesheets/reset.css';
+import '../stylesheets/login_form.css';
```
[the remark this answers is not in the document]
e91096: also i'm still seeing the vote form button style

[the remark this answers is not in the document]
chrisruenes1: thanks for catching this! I changed a class name and removed the duplicate files and I think this is fixed now

frontend/routes.jsx
chrisruenes1: yeah that is definitely unexpected behavior. I think for the sake of getting this merged in, I agree that it's fine for now and we can try to work on a solution when we're refactoring routing for this part of the app in general!

PR #75, opened 2018-09-27
Fix Mixpanel Error

brycemcd:
Errors in the console == bad. No errors == good.
This commit is, by definition, good.

## Prerequisite PRs
None

## Dependent PRs
None

## New ENV Vars
None

## Notes for Reviewer
Open it up in a browser. Marvel in the glorious lack of mixpanel errors.


app/views/layouts/application.html.erb
```diff
@@ -23,13 +23,19 @@
...
+      for(h=0;h<k.length;h++)e(d,k[h]);a._i.push([b,c,f])};a.__SV=1.2;b=e.createElement("script");b.type="text/javascript";b.async=!0;b.src="undefined"!==typeof MIXPANEL_CUSTOM_LIB_URL?MIXPANEL_CUSTOM_LIB_URL:"file:"===e.location.protocol&&"//cdn4.mxpnl.com/libs/mixpanel-2-latest.min.js".match(/^\/\//)?"https://cdn4.mxpnl.com/libs/mixpanel-2-latest.min.js":"//cdn4.mxpnl.com/libs/mixpanel-2-latest.min.js";c=e.getElementsByTagName("script")[0];c.parentNode.insertBefore(b,c)}})(document,window.mixpanel||[]);
+      mixpanel.init("<%= ENV.fetch('MIXPANEL_KEY') %>");</script><!-- end Mixpanel -->
```
e91096: why `ENV.fetch()` instead of `ENV[]`?

brycemcd: Mostly force of habit :) [Fetch will raise an exception if a sensible default is not supplied as a second argument](https://ruby-doc.org/core-2.2.0/Hash.html#method-i-fetch) and if we didn't have that env var set then I wanted our error-reporting tooling to make us aware of it. I can change it back to the not fetch version if you prefer. We have some other tooling that will make sure this env var is set and it would be more consistent with the rest of the code. Lemme know your preference please.

e91096:
Ahh gotcha! Very cool. As long as this doesn't mean the user might be
blocked bc of programmer error, I think this is fine!

PR #76, opened 2018-09-28
Add Mailchimp Importer

brycemcd:
Sometimes, users will sign up directly from MC (this is
the case before we've deployed user auth/auth for
example). We want to be able to run a script with a console
open and import all our MC users as constituents

[Trello Card](https://trello.com/c/P12O1AAy)

## Prerequisite PRs
None

## Dependent PRs
None

## New ENV Vars
None

## Notes for Reviewer
There's some code duplication from the Mailchimp class but it's on purpose. I think it's fine because it allows for either of those classes to be edited freely and it's just boiler plate from the Gibbon gem.


brycemcd:
Mother of pearl. Local tests are passing but failing in CI. :angry: I'll figure this out but please review the code in the meantime.

app/services/mailchimp_importer.rb
```diff
@@ -0,0 +1,73 @@
...
+  # notest
+  def all_members_in_list(list_id: ENV['MAILCHIMP_LIST_ID'])
+    @all_members_in_list ||= fetch_from_list(list_id)
+  end
+
+  def map_all_members_into_users(members_in_list: all_members_in_list)
+    # NOTE: I'd normally want to exit on failure so we wind up in a consistent
+    # state. This whole process is idempotent so we can pass on errors for now
+    #
+    # I also assume that this process is run manually with a piece of wetware
```
chrisruenes1: Oh man `wetware` is a *rough* phrase

app/services/mailchimp_importer.rb
```diff
@@ -0,0 +1,73 @@
...
+            email: cons[:email_address],
+            zip: mf[:ZIP_CODE],
+            first_name: mf[:FNAME],
+            last_name: mf[:LNAME],
+            congressional_district: mf[:DIST_NUM],
+            state: mf[:STATE],
+            user_token: mf[:USER_TOKEN],
+            us_citizen: mf[:US_CITIZEN] == 'Yes',
+          )
+          pw = u.send(:random_alphanum, 20)
```
chrisruenes1: what is the reason that we couldn't call this directly? As in `u.random_alphanum(20)`?

app/services/mailchimp_importer.rb
```diff
@@ -0,0 +1,73 @@
...
+      rescue StandardError => e
+        logger.error "failed to create users #{e}"
+        next
+      end
+    end.compact
+  end
+
+  # NOTE: the users is an array of user objects. The default value is _the result_
+  # of a function call, not a function itself
+  def persist_new_users(users: self.map_all_members_into_users())
```
chrisruenes1: I'm not familiar with this syntax...is it a rails5 thing?

app/services/mailchimp_importer.rb
```diff
@@ -0,0 +1,73 @@
+# imports our mailchimp list to our database in an idempotent process
+
+class MailchimpImporter
+  attr_reader :mailchimp, :logger
+
+  def initialize(debug: false, logger: Rails.logger)
+    if ENV['MAILCHIMP_API_KEY']
```
chrisruenes1: should we raise an error if this envvar doesn't exist? As far as I can tell from looking at this, it seems like it would raise a less clear error somewhere down the line if this conditional evaluated to false

app/services/mailchimp_importer.rb
brycemcd: It's because `random_alphanum` is a private method in `User`. You're right to pick up on this code smell. Normally, anytime you use `send`, you'd want to use `public_send` to be safe. In this case, properly factoring the code would probably mean bringing that method out into its own module where it could be included. In this case, I figure this import class will be used one time and then probably either deleted or re-modeled for some future use case. I went with the easiest implementation this time.

app/services/mailchimp_importer.rb
brycemcd: You mean the `method(keyword_argument: sensible_default_value)` syntax? This was new in Ruby 2.0 [Thoughbot wrote up a nice primer on it](https://robots.thoughtbot.com/ruby-2-keyword-arguments)

app/services/mailchimp_importer.rb
brycemcd: good point. The env var is defined as `required` in config/initializers/required_env_vars.rb so the app won't even boot unless that env var is set. I'll take the `if` statement out entirely.

app/services/mailchimp_importer.rb
chrisruenes1: Hmmm looking at this again, I think I was thrown off because:

1) i don’t _usually_ see method invocations in default arguments 
2) I’m not used to seeing a method that takes no arguments invoked explicitly (with the parentheses) in ruby. But I think that’s just a stylistic choice?

chrisruenes1:
@brycemd provided I’m right on my second look that the stuff I thought seemed unfamiliar was just stylistic, this LGTM!

PR #80, opened 2018-10-12
fix user sign in error message

e91096:
[Trello Card](https://trello.com/c/snytxDZ8/64-fe-users-shouldnt-see-email-pw-mismatch-error-message-if-login-failed-due-to-other-reasons)

## Notes for Reviewer
users shouldn't see email/pw mismatch error message if login failed due to other reasons


chrisruenes1:
@e91096 since my machine is in a state that I think could most aptly be described as “lean bricked”, I can’t Test this at the moment, but all of the code looks good to me! If you merge, I could make myself available at some point today to do a smoke test!

PR #84, opened 2018-10-17
Add docs: frontend framework

andrezafeu:
None

## Prerequisite PRs
None

## Dependent PRs
None

## New ENV Vars
None

## Notes for Reviewer
None


andrezafeu:
Any of the frontend framework options would work, really, and I would be happy with any of them.
I chose Bootstrap because I think it's the most popular and simple to use. Also, I've worked with the latest version and it has several improvements in the way that it looks and some new classes that can speed up our work.

andrezafeu:
Thanks for your comments!
@brycemcd All of them use a grid system. Foundation has just one break point for structural change. (The break point defines the viewport width at which the larger grid is put into action). Materialize has 3 breakpoints and Bootstrap 4.

@chrisruenes1 there are different ways to install Bootstrap (https://blog.logrocket.com/how-to-use-bootstrap-with-react-a354715d1121). I think that if we install it via CDN, it shouldn't have anything to do with the Rails version. Perhaps at your work they had it installed via the bootstrap gem?

I agree that they tend to turn styles into a black box, although the new Bootstrap version looks much prettier than the previous one. I think it really comes down to a trade-off between control over the design vs development speed + consistency across different devices. Which one should be prioritize?

I find CSS tasks extremely time consuming, and since as of right now we don't have a designer, I strong advocate for using a framework. I've worked with Bootstrap and Materialize and they are very easy to learn. Once it's set up, that's not much that can go wrong with it. And, we don't have to use it at every page ; )

Do you have other tools in mind that could help us in this phase that perhaps I'm not considering?

andrezafeu:
Oh oh, It looks like we are a bit divided over here ; )

@e91096 I think flexbox is a great tool to substitute the grid system, but I would also point out that bootstrap does a little bit more than that. I personally find the components (https://getbootstrap.com/docs/4.0/components/alerts/) super useful.

That being said, I think we should try to make a decision that everyone feels comfortable with! I've never worked in a front-end project that hasn't used a framework, so maybe I'm overestimating the time it would take to build the pages.

I feel comfortable working with or without using a framework. I lean towards using one just cause I think it's a better trade-off in terms of speed and consistency.

@brycemcd what's your opinion on the issues pointed out by @e91096 and @chrisruenes1 ? From a standpoint of someone that is creating the stories, establishing the deadlines, managing the resources, etc, do you still think it's a good trade-off?

andrezafeu:
Oh oh, It looks like we are a bit divided over here ; )

@e91096 I think flexbox is a great tool to substitute the grid system, but I would also point out that bootstrap does a little bit more than that. I personally find the components (https://getbootstrap.com/docs/4.0/components/alerts/) super useful.

That being said, I think we should try to make a decision that everyone feels comfortable with! I've never worked in a front-end project that hasn't used a framework, so maybe I'm overestimating the time it would take to build the pages.

I feel comfortable working with or without using a framework. I lean towards using one just cause I think it's a better trade-off in terms of speed and consistency.

@brycemcd what's your opinion on the issues pointed out by @e91096 and @chrisruenes1 ? From a standpoint of someone that is creating the stories, establishing the deadlines, managing the resources, etc, do you still think it's a good trade-off?

andrezafeu:
Oh oh, It looks like we are a bit divided over here ; )

@e91096 I think flexbox is a great tool to substitute the grid system, but I would also point out that bootstrap does a little bit more than that. I personally find the components (https://getbootstrap.com/docs/4.0/components/alerts/) super useful.

That being said, I think we should try to make a decision that everyone feels comfortable with! I've never worked in a front-end project that hasn't used a framework, so maybe I'm overestimating the time it would take to build the pages.

I feel comfortable working with or without using a framework. I lean towards using one just cause I think it's a better trade-off in terms of speed and consistency.

@brycemcd what's your opinion on the issues pointed out by @e91096 and @chrisruenes1 ? From a standpoint of someone that is creating the stories, establishing the deadlines, managing the resources, etc, do you still think it's a good trade-off?

brycemcd:
I do love a good debate :)

For the criticisms:

>  create a situation where styles are dependent on BE framework versioning choices
> - jean-Chris Ruenes, c. 2018

I think this is a case of strong coupling between your FE and BE at work. This should be strenuously avoided here and I think we have some good practices in place to avoid it. First, if we incorporate a FE framework, let's be careful not to add it as a gem. There are no ruby dependencies in our FE at the moment and we can simply use the CDN'd and minified version of Bootstrap and include it directly in the markup.

>  ... less important once we maybe have a designer who can help ...
> - jean-Chris Ruenes, c. 2018

I don't think I agree with this only insofar as a designer can help establish a design system. That system can be implemented using a framework. We can, for example, use Bootstrap's grid to put components of the design in the right place. Whomever is implementing the FE will still need to define the grid, color palette, and specific components for our design system. Bootstrap adds some useful components to prevent us from reinventing the wheel on those common elements.

> ...  I are fluent in any CSS frameworks, so there’d be some overhead cost in us having to learn ...
> - jean-Chris Ruenes, c. 2018

@chrisruenes1 do you learning the system has utility or is that time better spent creating those components in CSS ourselves?

> ... they tend to be heavy ...
> - jean-Esther, c. 2018

Can you more definitively define "heavy" please @e91096 ? I take this to mean size to download to the client. I can see it also meaning arduous coding to implement.

The size of the current minified Bootstrap framework to be downloaded by a client is 143KB according to my network inspector. IMHO, that seems worth it.

> ... sort of direct the way the html has to be written ...
> - jean-Esther, c. 2018

This is a good point. When I've used these in the past I've seen FE devs create a lot of extraneous markup. Is it worth it? My gut says a little extra markup is probably worth the savings we get in defining our own framework.

> ... and give websites a generic look
> - jean-Esther, c. 2018

Very true, though in our case at this moment, that might be a reasonable tradeoff. The alternative is to code && design everything from scratch ourselves.

---

My own $0.02 is that Bootstrap has a lot to offer. So long as we avoid coupling it to the backend :man_facepalming: , we get a lot of common components developed by some great UX/Product designers for free. None of us, as far as I know, have those skills and it would be nice to stand on their shoulders. 

You've all known me long enough to know that I don't like to think much :) Having a system that we can draw from to add new components to the UI as we're building out new features is helpful to prevent us from either creating them from scratch or having to look them up.

I also think it will help keep our design and code consistent across various areas of the product. Having Bootstrap incorporated will change the way we write CSS and see some positive changes:

* Our custom CSS will be specific to our product; no boilerplate
* It _should_ (but not required) create a way for us to organize our specific design components so they can be easily looked up and re-used. No spelunking through "votes.css," for example, looking for the way we style radio buttons.
* The total amount of CSS we write will be much less than if we had no framework at all. I like deleting code a lot; I like not writing unnecessary code even better :)

Those are my thoughts but I'd love to hear your reactions @andrezafeu , @chrisruenes1 and @e91096 !

andrezafeu:
@brycemcd also for me using Bootstrap is a good trade-off.

It is true also what you said about the designer role. We can implement an identity and still use a framework: they are not mutually exclusive. Also, we should consider the possibility that the designer will not necessarily be a web developer...

Regarding the loading speed I would also like to add that, if we decide to use CDN, Bootstrap will most likely be already cached by the user since so many websites use it.

@e91096 and @chrisruenes1 , any new thoughts about this?

chrisruenes1:
Somehow I missed all these updated until today! Sorry for my delay responding. Ok, I think I’m convinced by these arguments! Especially 1) that we can install bootstrap using a CDN and therefore have it be entirely independent from the BE and 2) that it can be *a* tool for us without mandating that it is *the* tool for us. @e91096 what do you think??

PR #85, opened 2018-10-18
Logs user in when submitting a vote

brycemcd:
For a _very_ temporary period of time. We'll use our email system to authenticate users. I want to stress that this is _not_ good practice but rather a convenience compromise for user experience while we migrate our users completely off of Google Forms

[Trello Card](https://trello.com/c/0TwPNWQc)

## Notes for Reviewer
Submit a vote with a valid value in the `userToken` parameter in the query string. Check local storage and be amazed by the token that exists.


chrisruenes1:
@brycemcd generally this LGTM! I think I’m not totally up-to-speed with auth in our app, so could you just clarify what the precondition is for reverting this temp change? Otherwise, I say 🚢!!

brycemcd:
@chrisruenes1 after our most active alpha testers roll through and get authenticated, we'll role this back.

PR #86, opened 2018-10-18
OPPREF - update vote_queuer for Sausage changes

brycemcd:
The API for sausage is changing and since that app is the only
consumer of vote_queuer, I can make it simple and change this
code too.

## Prerequisite PRs
None

## Dependent PRs
None

## New ENV Vars
None

## Notes for Reviewer

Spin it up in a console and marvel in the beauty of the message


lib/tasks/vote_queue.rake
```diff
@@ -0,0 +1,11 @@
+namespace :vote_queue do
+  desc "For every user, look up votes in the last 48 hours and write to SQS"
+  task write_recent_votes_to_queue: :environment do
+    User.in_batches.each do |user_batch|
```
e91096: 😮 `in_batches` is a thing?! do the batch operations run concurrently?

brycemcd: This is Ruby. Nothing is concurrent :trollface:

`in_batches` prevents loading the whole user table into memory at the same time. It's handy for when our user table is several gigabytes in size and we want to work with KB or MB at a time.

PR #90, opened 2018-10-22
BE - Update validations and strong params

brycemcd:
Allows for password confirmation and removes congressional
district requirement. Sets -99 as default congressional district
so we can easily query for a known system state (as opposed to null
which may be an unknown system error)

[Trello Card](https://trello.com/c/6IjBOaP6)

## Prerequisite PRs
None

## Dependent PRs
None

## New ENV Vars
None

## Notes for Reviewer

Play around with the curl command in the docs. Try and get errors as expected


chrisruenes1:
@brycemcd 👍 ! I am train-reviewing so I wasn’t able to test, but I like the -99 searchable solution and the code all checks out to me. A question more of curiosity than critique: did you consider making this a Postgres default value, and if so, why did you go with the before_validation hook instead?

brycemcd:
@chrisruenes1 I did consider making -99 a default value but I thought the application should handle error codes. If we have a different acceptable error state, for example an automatic district calculator errors and needs manual intervention, we can set a -98 in that column and select only those error states in a query.

PR #100, opened 2018-11-18
hitting 'Enter' submits login/registration forms

e91096:
[Trello Card](https://trello.com/c/F9ZlHf6F/67-fe-hitting-enter-should-submit-forms)

## Notes for Reviewer
you should now be able to enter your username/password and then hit enter to log in! 
also works in the registration form.


frontend/components/RegistrationForm.jsx
```diff
@@ -236,6 +240,7 @@ class RegistrationForm extends Component {
           id="US-citizen"
           label="U.S. Citizen*"
           handleChange={isChecked => this.handleInputChange('USCitizen', isChecked)}
+          handleEnter={this.handleSubmit}
```
brycemcd: There is a lot of duplication here. Is there a way to add this handleEnter attribute to all input elements?

brycemcd:
Aside from the one comment, LGTM

frontend/components/RegistrationForm.jsx
e91096: agreed; i didn't see a way of doing it in less lines that would be preferable

PR #102, opened 2018-11-30
Add before_action to overwrite user token in campaigns controller

chrisruenes1:
After merging in the MR that overwrites the user token if a user is authenticated (https://github.com/PocketLobby/constituents/pull/95), the votes in the `show` endpoint for campaigns was falling out of sync. This was because the `votes_controller` was overwriting the user token when CRUDing votes, but the `campaigns_controller` was not, so the token stored in local storage was being used to fetch votes.

## Notes for Reviewer

Register a user, then submit votes for the campaign you're redirected to. Hit the back button after the campaign redirect, and the votes you just made should be loaded into the frontend.


chrisruenes1:
@brycemcd I went for the most straightforward solution here, but do you think it would make sense to abstract that callback into the `application_controller`? Or would that present a higher security flaw?

PR #104, opened 2018-11-30
fix current campaign reducer

e91096:
[Trello Card](https://trello.com/c/a3IsOLlT/102-bug-fe-user-not-redirected-to-current-campaign-after-login)

context in trello card

## Notes for Reviewer
user should now be redirected to the current campaign after login/registration


brycemcd:
I'm not sure we want the new constituent going to the current campaign. Rather, we have an onboarding campaign set up for him/her. Could we send them to that campaign instead? We could identify the campaign id by using an ENV variable.

brycemcd:
@e91096 if the constituent isn't redirected anywhere after registration/login right now, merge this in and we'll create a cleanup ticket to redirect to `current` campaign after login and `onboarding` campaign after registration. After six days of brokeness let's get to not-broken and then less-broken.

e91096:
yup i'll merge this now! the onboarding campaign is a great idea and i think should be its own ticket

PR #109, opened 2019-01-14
Save legislator votes

chrisruenes1:


app/models/legislator_vote.rb
```diff
@@ -0,0 +1,6 @@
+class LegislatorVote < ApplicationRecord
+  validates_inclusion_of :response, in: ["yes", "no", "abstain", "speaker"]
```
e91096: no `%w` here?

app/controllers/api/bill_summaries_controller.rb
```diff
@@ -11,6 +11,7 @@ def index
   def create
     bill_summary = BillSummary.new(permitted_params)
     if bill_summary.save
+      LegislatorVoteFetcher.fetch(bill_summary.bill_number)
```
e91096: if this fetch fails, do we want to roll back the bill summary save?

app/models/legislator_vote.rb
```diff
@@ -0,0 +1,6 @@
...
+  validates_inclusion_of :vote_type, in: ["voice_vote", "roll_call_vote"]
```
e91096: also here

PR #112, opened 2019-01-14
add ensureLoggedIn check to routing

e91096:


chrisruenes1:
Awesome. This looks great overall. Maybe we could build a separate unauthorized component so that users don't get bounced to the success page if they are not able to access the url? I think we could merge this in and work on that later!

PR #114, opened 2019-02-03
Update legislator_vote_fetcher to use govinfo for voice vote data

chrisruenes1:
Our use of the pro publica `versions` data and a whitelist of constants to try to distinguish between a voice vote and no vote was pretty brittle.

Govinfo returns xml data on bills that includes `actionCode` data, which map onto explicit states that a bill can be in. We'll want this info eventually for when we try to do some bill status stuff locally, but two of them are about explicit passage in the house and senate (8000 and 17000) respectively. You can see a full list of the status codes under section 3 of this readme https://github.com/usgpo/bill-status/blob/master/BILLSTATUS-XML_User_User-Guide.md#3.-Action-Code-Element-Possible-Values

Basically, follow the same strategy as before, except the criteria for something being considered a voice vote is it having the appropriate status code in its list of actions.

## Notes for Reviewer

Try uploading a couple bills and make sure the results map to what you expect.  I tested with this one https://www.congress.gov/bill/116th-congress/house-joint-resolution/31/actions which caused a problem in our staging environment.


app/services/legislator_vote_fetcher.rb
```diff
@@ -84,27 +74,34 @@ def handle_voice_vote_for_both_chambers
...
-      when "Senate"
-        versions.find {|version| SENATE_VOICE_VOTE_STATUSES.include?(version["status"])}
+    bill_type = @bill_number.match(/[a-z]*/).to_s
+    # govinfo provides status codes for the bill's progress, which is more stable than
+    # trying to search through the versions like we were in the pro publica data
+    url = "https://www.govinfo.gov/bulkdata/BILLSTATUS/#{ENV["CONGRESS"]}/#{bill_type}/BILLSTATUS-#{ENV["CONGRESS"]}#{@bill_number}.xml"
+    xml_result = RestClient.get(url)
+    result_hash = Hash.from_xml(Nokogiri::XML(xml_result).to_s)
+    bill_status = result_hash["billStatus"]
+    unless bill_status.empty?
```
e91096: is `result_hash["billStatus"]` always defined? (ie we'd never be calling `.empty?` on `nil` here would we?)

PR #115, opened 2019-02-11
Refactor votes page and fix 115 congress

chrisruenes1:

## Prerequisite PRs
#114 

## Notes for Reviewer

I think the best way to test this is to call `rake legislators:populate[115]` after migrating, to get a list a of legislators from the 115th congress in the database. Then, try to add some bill summaries from both 115 and 116, across two or three different campaigns.

The most egregious bug was coming from North Dakota, where the current junior senator was a representative in the last congress. Pretty much all parts of our app hated that, so there's a bunch of places the logic had to be updated. But if you change the state and district of your user to `ND:1`, then you should be able to see:

1) the matches page broken up by campaign, with a bad name per campaign (that needs to be fixed)
2) a different set of legislators for campaigns from the 115th congress than from the 116th congress


Gemfile
```diff
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-ruby '2.5.3'
```
chrisruenes1: SO this is branched off of production, which I know is not the best git practice, but I figured we could resolve it after this stuff is out. I upgraded to ruby 2.6.0 on prod because that is where Elastic Beanstalk is at now

app/models/bill_summary.rb
```diff
@@ -14,7 +14,7 @@ class BillSummary < ApplicationRecord
   scope :campaign_sorted, -> { order(updated_at: :asc) }
 
   def vote_types_for(chamber)
-    legislator_votes.joins(:legislator).where(legislators: { chamber: chamber }).first.try(:vote_type)
+    legislator_votes.joins(:legislator).where(legislators: { chamber: chamber, congress_session: self.congress_session }).first.try(:vote_type)
```
chrisruenes1: at some point I think I'd like to change this to `congressional_session`, because `congress_session` sounds really terrible

app/services/legislator_vote_fetcher.rb
```diff
@@ -7,25 +7,17 @@ class LegislatorVoteFetcher
...
-    "Enrolled as Agreed to or Passed by Both House and Senate",
-    "Considered and Passed by Senate",
-    "Engrossed as Agreed to or Passed by Senate"
-  ]
-
-  HOUSE_VOICE_VOTE_STATUSES = [
-    "Enrolled as Agreed to or Passed by Both House and Senate",
-    "Considered and Passed by House",
-    "Engrossed as Agreed to or Passed by House"
-  ]
```
chrisruenes1: some of these changes are from the other PR, 114. I can target that if you want!

chrisruenes1: The rest of the changes in this file make it possible to pass in a `congress`, in case the bill comes from a different congress than the current environment's congress

app/services/mailchimp.rb
```diff
@@ -2,32 +2,31 @@ class Mailchimp
   attr_reader :mailchimp
 
   def initialize
-    # TODO comment all this back in once we've restored mailchimp
-    # @mailchimp = Gibbon::Request.new(api_key: ENV['MAILCHIMP_API_KEY'],
-    #                                  symbolize_keys:  true)
-    # # optional: pass debug:true to @mailchimp
-    # @mailchimp.timeout = 30
-    # @mailchimp.open_timeout = 30
+    @mailchimp = Gibbon::Request.new(api_key: ENV['MAILCHIMP_API_KEY'],
```
chrisruenes1: we shouldn't deploy this to next yet without setting up a separate mailchimp list, but this is live on prod again

app/services/vote_matcher.rb
```diff
@@ -18,7 +18,8 @@ def as_json
         user_id: vote.user_id,
```
chrisruenes1: changes here allow us to group by campaign on frontend and to organize legislator votes by congress session

app/views/layouts/application.html.erb
```diff
@@ -10,10 +10,32 @@
 
     <%= stylesheet_link_tag    'application', media: 'all', 'data-turbolinks-track': 'reload' %>
     <%= javascript_include_tag 'application' %>
+    <link rel="stylesheet" href="https://unpkg.com/react-select/dist/react-select.css">
```
chrisruenes1: wait did we get rid of this?

app/views/layouts/application.html.erb
```diff
@@ -10,10 +10,32 @@
...
     <link href="https://fonts.googleapis.com/css?family=Open+Sans:400,700,800" rel="stylesheet">
+
```
chrisruenes1: below is all stuff I deleted just to get the thing out to staging but that we want in production again

